PR #553, opened 2015-09-05
Issue 542 and 549

majuca:
Issue 542: the function validate return true in any case but should return false if this.isValid equal false.
Issue 549: the condition of the criteria cannot be reseted when empty value


majuca:
Hi,

According to the issu 549, there are an error in the new commit. It must not be

if(empty($value)) {
// reset the condition
$this->_condition = '';
return;
}

but:

if(empty($value)) {
// reset the condition
$this->_condition = null;
return;
}

Issue #561, opened 2016-01-14
findByPk with null


majuca:
Hi, we have an exception when w try:

```
MyModel::finder()->findByPk(null);
```

The problem happens on the file TDataGatewayCommand.php, function findByPk, line 163:

```
list($where, $parameters) = $this->getPrimaryKeyCondition((array)$keys);
```
indeed, (array)null give a zero size array and in the function getPrimaryKeyCondition the following line crash when execute $values[$i++]:

```
$bindings[$key] = isset($values[$key])?$values[$key]:$values[$i++];
```

ctrlaltca:
As far as i know the primary key can't be nullable. Is there any any DBMS that permits this?
In your opinion, should we return anything when ```findByPk(null);``` is called, or just error out?

majuca:
I am not sure, but I think that a primary key could not be null, but for me, findByPk with a null sould not throw an exception but just return null like when no record is found. What do you think?

ctrlaltca:
I agree with you: http://pradosoft.github.io/docs/manual/class-TActiveRecord.html#_findByPk
Returning null is probably the best behavior here.

majuca:
ok, i can do a PR if you want?

ctrlaltca:
> ok, i can do a PR if you want?

:+1:

ctrlaltca:
Fixed by #564

PR #564, opened 2016-02-02
Return null when keys is null and don't throw an exception

majuca:


ctrlaltca:
I suppose this fixes #561 ?

majuca:
Hi, yes this fixed #561

Issue #559, opened 2015-12-22
Prado, php7 and memcache->memcached


majuca:
Hi,

We tried our application with php7 and we are very happy. Our application is 40-60% faster with php7 and everything is working like a charm without any modification. The only problem we have is memcache. With php7, it is not memcache but memcached. So I look on TMemCache and it is working by modifying the following code:

```
public function init($config)
{
		if(!extension_loaded('memcached')) 
			throw new TConfigurationException('memcache_extension_required');
		$this->_cache=new Memcached;
```

ctrlaltca:
Thank you for reporting back on php7; we tested it, too, and it seems to run fine and faster :)
About memcache and memcached, those are different PECL extensions.
Memcached has already been ported to php7: https://github.com/php-memcached-dev/php-memcached/tree/php7 , while memcache won't probably see a php7 release.

Talking about backwards compatibility: Memcached supports all the php version supported by Prado, but since it requires a different PECL extension to work, sowe should either:
 * create a new ```TMemCached``` component and keep the old one, or
 * make the old control use the "new" extension and detail this change in the UPGRADE notes

Since we're about to push a new release, we could go with the second option.
I'll keep this ticket open until the change gets in to let others comment.

majuca:
A third solution could be to have a new setting to choose which PECL extension you want to use with something lile:
```
<module id="cache" class="System.Caching.TMemCache" Host="localhost" Port="11211" Extension="memcache" />
```
or
```
<module id="cache" class="System.Caching.TMemCache" Host="localhost" Port="11211" Extension="memcached" />
```

ctrlaltca:
Yep, i saw the Yii extension is doing exactly the same: https://github.com/yiisoft/yii/blob/master/framework/caching/CMemCache.php
We are currently exposing a couple of options (related to ```http://php.net/manual/en/memcache.setcompressthreshold.php```) that won't work with the new extension; also, persistent connections are handled differently between the two extensions. Looking at the Yii code, persistent connections are not working when using memcached.

ctrlaltca:
By the way, it seems like they got it fixed in yii2: https://github.com/yiisoft/yii2/blob/master/framework/caching/MemCache.php#L222

majuca:
I will try to propose something by a PR.

ctrlaltca:
> I will try to propose something by a PR.

:+1:

majuca:
Thank @ctrlaltca for the merge. This issue could be closed.

PR #565, opened 2016-02-11
Issue #559, add support for memcached

majuca:
According to the discussion in the issue #559, I implemented the support of memcached


ctrlaltca:
Thank you

PR #570, opened 2016-02-26
Replace .page by .html

yoisar:
I always have to do this replacement . This is because I always work with HTML tools like Dreamweaver. If there is any way to configure the application. For example, a parameter that is templateExtention apalicacion or something


javalizard:
The 3.2 branch isn't the 3.3 branch.  There are changes in here that would upgrade the 3.2 branch to 3.3 making it NOT the 3.2 branch; such as defining prado versions to 3.3, etc.

This isn't to say 3.2 shouldn't me maintained.  3.2 can be maintained if necessary but not by making it 3.3.

The easiest method to redefine your own ".page" extension to ".html" is to duplicate the TPageService file into a THTMLService.  Then simply change the PAGE_FILE_EXT and reroute your application services configuration.

This is a another great example of Behaviors where we could use Dynamic Events to filter the Default value from ".page" to ".html" by replacing `self::PAGE_FILE_EXT` with `$this->dyDefaultPageFileExt(self::PAGE_FILE_EXT)`
Though there may be a reasons for leaving this as a direct constant.  1. Security (though if someone can add a behavior to the page service object without proper authority there are far greater concerns for the sysadmin!)    2. no great method for debugging behaviors and looking at what objects have what behaviors, yet.

composer.json
```diff
@@ -37,17 +40,17 @@
...
 		"phpunit/phpunit-story": "*",
-		"phing/phing" : "2.4.14",
-		"apigen/apigen": "2.*",
+		"phing/phing" : "2.*",
+		"apigen/apigen": "4.*",
 		"satooshi/php-coveralls": "dev-master"
 	},
 	"provide" : {
-		"pradosoft/prado" : "3.2.3"
+		"pradosoft/prado" : "3.3.0"
```
javalizard: Why put a 3.3 tag in the 3.2 branch?

javalizard:
I don't mean to be a stickler....   It looks like there are some great changes.  Let's be thorough.

javalizard:
Should we be upgrading the PHP version for the 3.2 branch?  Is backward PHP compatibility important for the 3.2 branch?

ctrlaltca:
Closing by now. Please rebase it to master and re-apply for a new PR, so that we can actually see what you're trying to commit. Thank you.

yoisar:
Ohh!! man It was late at night. I made changes but apparently my computer was half frozen and climbed the wrong change . I am so sorry. So much!

yoisar:
I have no permission to pull request . As I can get these permits?

ctrlaltca:
Everyone has permission to open a new pull request (as you did 9 hours ago). Do you see a specific error message somewhere?

yoisar:
Ok, I have some components I want To share. And some changes. I try to Sync, and i have no permission

ctrlaltca:
You can't commit directly into prado's master. You need to fork it, open a new branch containing the changes and then propose to get them merged using a pull request.

yoisar:
ok thanks I´ll do it

yoisar:
nop. Syc Falied to push local changes! :(

PR #579, opened 2016-03-30
Added TClientStyleSheet::PradoStyles

### LINKED ISSUE #562: Suggesting new feature "TClientStyleSheet"
Good afternoon,

I would like to make a suggestion for a feature which would have a similar "TClientScript" behavior but in this case it would not render a JS file if a CSS file as an idea for the name would be "TClientStyleSheet" and that likewise allow register a package of CSS files.

```html
<com:TClientStyleSheet PradoStyleSheet="bootstrap" />
<com:TClientStyleSheet StyleSheetUrl=<%~ test.css %> />
```

ctrlaltca:
Fix #562

### DIFF
--- a/HISTORY
+++ b/HISTORY
@@ -15,6 +15,7 @@ ENH: Intercept fatal errors using register_shutdown_function (ctrlaltca)
 ENH: Activecontrols: avoid updating client side if the value didn't change (ctrlaltca)
 ENH: Added TReCaptcha2 / TReCaptcha2Validator (camilohaze)
 BUG: Fix callback update of TActiveCheckBoxList when initially empty (LCSKJ)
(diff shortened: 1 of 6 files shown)


ctrlaltca:
@camilohaze can you check this? Example usage:
```html
<com:TStyleSheet PradoStyles="bootstrap, jquery.ui.progressbar" />
```

camilohaze:
It works perfectly.

ctrlaltca:
> It is necessary to force the user to use the StyleSheet control whenever you include within a THead?

Sorry, i don't get the exact meaning of the question.
In order to use TStyleSheet you need a THead, because the ```<link>``` and ```<style>``` will be rendered there.
You're not forced in using a TStyleSheet anyway; you can use plain html tags, just prado won't know anything about them.

PR #587, opened 2016-04-12
PHP7 compatibility fixes

emkael:
Hi,

I've run [PHP7 Compatibility Checker](https://github.com/sstalle/php7cc) through the framework code and fixed some issues that were detected:

 * use of removed `e` regex modifier in `TextHighlighter` library (upstream does not provide a fix for that, in fact, the code hasn't been updated for years)
 * PHP4-style constructors in `Gettext` classes
 * removed the `MySQL` message source in `I18N`, since `mysql_*` functions are gone from PHP7

That last one commit may be a bit too much (maybe a `mysqli_*` rewrite would be more suitable?), I can see that `MessageSource_MySQL` has been deprecated for nearly 8 years, though.

FYI: all the other issues detected by php7cc do not seem to be dangerous, at first glance.
There are lots of warning about changed behavior of `func_get_args`, some "possible array element creation during by-reference assignment" warnings and a single warning to make sure that custom session save handler works properly (it does, and it's not registered unless a developer explicitly overrides `THttpSession` methods in a subclass).

Best wishes,
Michal K.


ctrlaltca:
Nice, thank you. I'm fine with dropping the long-time deprecated MessageSource_MySQL.
By the way i suppose a lot of more code is still using the mysql driver.

PR #591, opened 2016-05-08
Support for hyphenated attributes via <prop:*></prop:*> template syntax

emkael:
Commit ad5c394d39c35b48b7cfc4384288e53d268d0aab added to v3.2.* the ability to define, among other things, HTML5 data-* attributes via `Attributes.data-custom-data="value"` syntax in template controls.

This changeset should extend that functionality to support it within `<prop:*></prop:*>` template tags, too.
However, I have to admit that the piece of RegEx that's responsible for template parsing terrifies me a bit.

Also, after rebuilding pradolite.php, I've noticed that the recent TStyle constructor change wasn't compiled into it. I don't know what's the accepted procedure for pradolite.php in pull requests, so I apologize if adding it was unnecessary.

Michal K.


ctrlaltca:
Looks fine to me 👍 
There's no written procedure about when pradolite should be rebuilt, but i think that including it in an open PR is fine until it doesn't create merge conflicts with other PRs.

PR #622, opened 2016-12-17
Remove mcrypt, add OpenSSL support.

LCSKJ:
I changed the TSecurityManager to use OpenSSL functions instead of the deprecated mcrypt ones.
There are some things to mention:
- the methods marked as deprecated since Prado 3.2.1 were removed (i guess that is reasonable).
- default hash algorithm was changed to sha256 since md5 and sha1 are supposed to be vulnerable.
- default crypt algorithm was changed to aes-256-cbc from OpenSSL.
- since mcrypt used a rijndael-256 cipher as default which has no equivalent in OpenSSL, there may be undecryptable data after upgrading.


ctrlaltca:
Looks good, with a few minor things to address:
 * typo "OpenSLL" at line 42
 * The `TSecurityManagerValidationMode.php` class file can be deleted as it's unused.
 * Add a note about the need of OpenSSL extension in `UPGRADE`
 * in `composer.json` remove "ext-mcrypt" and add "ext-openssl"

LCSKJ:
Addressed your points in forked repo, commit was auto-added to pull request. If it's OK like this, feel free to merge.

ctrlaltca:
Perfect, thank you! 👍

PR #639, opened 2017-03-08
Fix #638

### LINKED ISSUE #638: Selecting item in TJuiAutoComplete after validation error
I added the following code to that [`Home.page`](http://www.pradoframework.net/demos/quickstart/?page=ViewSource&path=/JuiControls/Samples/TJuiAutoComplete/Home.page).
```html
<com:TTextBox ID="txtDep" />
<com:TRequiredFieldValidator ValidationGroup="val" Display="Dynamic" FocusOnError="true" ControlToValidate="txtDep" Text="Fill the textbox" />
<com:TButton Text="send" ValidationGroup="val" />                
```
When the page is loaded and you FIRST click on the button, the `JuiAutoComplete` don't work again even after requirement satisfy.

ctrlaltca:
When a validator with `FocusOnError=true` triggers, don’t call `Prado.Element.focus()` on the element but just focus it.
Currently the `Prado.Element.focus()` only works on ajax callbacks and it will probably need some rework.

### DIFF
--- a/framework/Web/Javascripts/source/prado/prado.js
+++ b/framework/Web/Javascripts/source/prado/prado.js
@@ -446,11 +446,14 @@ Prado.Element =
 
 	focus : function(element)
 	{
-		jQuery(document).ajaxStop(function () {
+		if(jQuery.active > 0)
+		{
 			setTimeout(function(){
 				jQuery("#"+element).focus();
 			}, 100);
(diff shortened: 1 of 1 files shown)


ctrlaltca:
Updated: properly fix Prado.Element.focus()

Issue #662, opened 2018-02-07
External inks to API manual doesn't work anymore


ctrlaltca:
API manual used to create class pages named after the lone class name, eg:
```
http://pradosoft.github.io/docs/manual/class-TDropDownList.html
```

Now the API manual is generating pages using the full namespaced class name:
```
http://pradosoft.github.io/docs/manual/class-Prado.Web.UI.WebControls.TDropDownList.html
```

Links from the quickstart tutorial and error messages are currently broken.

ctrlaltca:
Error messages fixed in #664 
Quickstart tutorial fixed in https://github.com/pradosoft/prado-demos/commit/7debb8042ff079e900a954bcc8eff9ce31069f49

PR #664, opened 2018-02-09
Fix internal link to API (issue #662)

frkinta:


frkinta:
Right. I didn't take this case in account. 
I modified the function to catch the generated Exception.

ctrlaltca:
Cool! thank you!

Issue #696, opened 2019-07-18
SQL error when trying to disable PDO::ATTR_EMULATE_PREPARES option for MySQL


ganiuszka:
Hello PRADO Group,

When I try to disable PDO::ATTR_EMULATE_PREPARES option in my application by:

$db_connection->setAttribute(PDO::ATTR_EMULATE_PREPARES, false);

then I receive an error from PRADO (and from MySQL server):

Internal error. TDbCommand failed to prepare the SQL statement "SHOW FULL TABLES LIKE :table": SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '?' at line 1

I have been trying to check what is going on and I noticed that MySQL native prepared statements feature doesn't support 'SHOW TABLES' syntax. As you can see 'SHOW TABLES' is not included in supported requests for prepared statements:

https://dev.mysql.com/doc/refman/5.7/en/sql-syntax-prepared-statements.html#idm140594804143440

It works only when emulation of prepared statements is enabled in PDO. This emulation is enabled in PRADO by default in TDbConnection::open() method but without any way to disable it due to the above error.

I would like to propose to allow setting PDO::ATTR_EMULATE_PREPARES to false in applications that use PRADO. In my case I develop application that supports MySQL, PostgreSQL and SQLite databases. In case MySQL without disabling PDO::ATTR_EMULATE_PREPARES all integer values returned from MySQL are strings instead of integers.

From the above reason I would like to propose to not use prepared statements in TMysqlMetaData::getIsView() method that will allow to disable PDO::ATTR_EMULATE_PREPARES in user applications if they need to do that.

I prepared simple patch that fixes this issue. I would like to ask you about considering using this patch. It is done for PRADO version 4.0.2.

Thank you in advance for your help.

Best regards,
Marcin Haba (gani)

[fix_disabling_pdo_attr_emulate_prepares.patch.txt](https://github.com/pradosoft/prado/files/3408133/fix_disabling_pdo_attr_emulate_prepares.patch.txt)

ctrlaltca:
Please open a proper PR with the patch.

ganiuszka:
Hello The PRADO Group,

I let you know that I have just opened the pull request.

Best regards,
Marcin Haba (gani)

ctrlaltca:
Fixed in #697, thank you!

PR #697, opened 2019-07-19
Fix SQL error when trying to disable PDO::ATTR_EMULATE_PREPARES optio…

ganiuszka:
Hello The PRADO Group,

I would like to propose fix to issue #696 It enables to set PDO::ATTR_EMULATE_PREPARES to false in user application without experiencing SQL error.

Thank you in advance for considering this change.

Best regards,
Marcin Haba (gani)


ctrlaltca:
The change looks fine, the same syntax is already in use in other methods (eg. `getConstraintKeys()`).
The only remaining method using prepared statements is `getForeignConstraints()`, but it should not be problematic since it executes a simple `SELECT`

ctrlaltca:
Fix merged and backported to 4.0, thank you!

PR #720, opened 2020-06-12
Fix usage of incorrect private property

qysp:
An incorrect private property in the existence check for the crypt algorithm has been used which always leads to an error when setting the algorithm.


ctrlaltca:
Nice catch, thank you!

Issue #488, opened 2013-10-14
TApplication::onInitComplete


javalizard:
We propose a new event in TApplication called onInitComplete.   onInitComplete is best called in initApplication before starting to process any services.  In this way, services may be added dynamically by any modules.  This is the method for doing any post initialization clean up.

This method is fundamental to application initialization of prado-cli commands.  onBeginRequest is NOT called during CLI for these requests.  onBeginRequest has been the event that developers are told to bind upon for any module post-initialization.  During CLI initialization, all modules requiring onBeginRequest are not properly initialized.

Of course, PRADO documentation (pictures and text) would need to be updated.

Like this:
TApplication.php

```
protected function initApplication()
{
    Prado::trace('Initializing application','System.TApplication');

    if($this->getConfigurationFile()!==null)
    {
        $_cacheFile=$this->getRuntimePath().DIRECTORY_SEPARATOR.TApplication::CONFIGCACHE_FILE;
        if($_cacheFile===null || @filemtime($_cacheFile)<filemtime($this->getConfigurationFile()))
        {
            $config=new TApplicationConfiguration;
            $config->loadFromFile($this->getConfigurationFile());
            if($_cacheFile!==null)
                file_put_contents($_cacheFile,Prado::serialize($config),LOCK_EX);
        }
        else
            $config=Prado::unserialize(file_get_contents($_cacheFile));

        $this->applyConfiguration($config,false);
    }

    $this->onInitComplete();

    if(($serviceID=$this->getRequest()->resolveRequest(array_keys($this->_services)))===null)
        $serviceID=$this->getPageServiceID();

    $this->startService($serviceID);
}

/**
 * Raises onInitComplete event.
 * At the time when this method is invoked, application modules are loaded, 
 * and the application is about to start processing the user request.
 */
public function onInitComplete()
{
    $this->raiseEvent('onInitComplete',$this,null);
}
```

belisoful:
Submitted a Pull Request for resolution of this enhancement.  I would like to call this a bug because Shell applications have no place for post initialize completion except for this.

the graphic on http://www.pradoframework.net/demos/quickstart/?page=Fundamentals.Applications needs to be updated too.

ctrlaltca:
fixed in #751

Issue #339, opened 2013-09-14
TComponent has no constructor- All sub-classes should call parent::__construct in their constructor


ctrlaltca:
_From [google...@pcforum.hu](https://code.google.com/u/112363926660835410968/) on June 25, 2011 07:53:45_

The TComponent in the current trunk has no constructor, thus descendant classes can't call parent::__construct() to preserve a proper constructor chain, regardless of actual inheritance hierarchy. This is bad practice.

Also TPageNotFoundEventParameter tries to call parent::__construct() in its own constructor, but since neither its direct parent (TEventParameter), nor the latter's parent (TComponent) define any, it fails with a PHP error. Triggered when you try to open a non-existing page.

_Original issue: http://code.google.com/p/prado3/issues/detail?id=338_

belisoful:
I am unifying the __construct methods.  There are many objects that wouldn't get their autoGlobalListen.  Testing this is a real pain, but it's happening.

belisoful:
With these commits, this issue can be put to rest.

ctrlaltca:
Fixed in #746

PR #732, opened 2021-03-25
#488 added onInitComplete for CLI application module initialization

belisoful:
This is an addition-enhancement that will help CLI applications initialize after the modules are loaded up.  It replaces OnBeginRequest as the first application event.  Shell applications do not have an application event stack like normal applications... but they still should have post module initialization.

Use Case, A Module that uses the the database module may want to do its database work in onInitComplete if, in case, the database module is loaded after the A Module.


framework/TApplication.php
```diff
@@ -1023,6 +1024,7 @@ public function applyConfiguration($config, $withinService = false)
 	 * Loads configuration and initializes application.
 	 * Configuration file will be read and parsed (if a valid cached version exists,
 	 * it will be used instead). Then, modules are created and initialized;
+	 * Finalize any modules by calling onInitComplete, mainly for CLI initialization.
```
ctrlaltca, line 1027: Please reword this; onInitComplete is an event, so the use doesn't "call" it.
Something like "module initialization can be finalized in the onInitComplete event, this is particularly useful for CLI applications"

framework/TApplication.php
```diff
@@ -1104,6 +1108,20 @@ public function onError($param)
 		$this->getErrorHandler()->handleError($this, $param);
 	}
 
+	/**
+	 * Raises onInitComplete event.
+	 * At the time when this method is invoked, application modules are loaded,
+	 * user request is resolved and the corresponding service is loaded and
+	 * initialized. The application is about to start processing the user
```
ctrlaltca, line 1115: At this step services are loaded  but not initialized yet: they have no ID assigned and their still have to load their config.

framework/TApplication.php
```diff
@@ -1104,6 +1108,20 @@ public function onError($param)
...
+	 * request.  This call is important for Shell applications that do not have
+	 * a web service stack.  This is the first and last call for finalization
+	 * of any loaded modules in Shell mode.
```
ctrlaltca, line 1118: Sorry for being this nitpick, but I find this a bit confusing for new users,  can you please reword this to make it more propositive, something like "In this event modules can perform any needed late initialization before the service starts to process the request" or something similar?

ctrlaltca (changes requested):
Hi, and first thank you for your PR.
The change itself looks fine, please have a look at the few possible improvements on the comments.

belisoful:
Yeah.  I'm not exactly great at the commenting thing yet.  Getting there.  You aren't Nitpicking...  Your comments are exactly what the code comments need!  I'm looking for feedback, so THANK YOU SO MUCH!!!   I'l make the changes and get back to you shortly.

belisoful:
My humble apologies, I'm beginning to think that onInitComplete should be after the service initialization.  Your thoughts?

belisoful:
BTW, I cannot describe how much I (and we) appreciate your upkeep of PRADO all this time.

ctrlaltca:
> My humble apologies, I'm beginning to think that onInitComplete should be after the service initialization. Your thoughts?

Maybe we could just make `TApplication::_steps` protected and override it in `TShellApplication::_steps` providing a few common steps that can be usable on a CLI application. Then `TShellApplication::run()` will just loop on the events like `TApplication::run()` already does.
What do you think of this approach?

belisoful:
From the standpoint of shell user input, looping on (a reduced number of) steps may be useful.  if the CLI is processing single commands, looping wouldn't be needed.  The CLI doesn't exactly have requests in the same way that the web app/service does.  unifying these steps between these differing app types would be of great help to modules. but does it make sense?  I'll have to look into the Shell App and what it's functionality is actually doing.

belisoful:
I have a whole slew of new functionality for Prado in mind.  I am writing tickets to address the short comings, their fixes, and hopefully quickly, checkins.

belisoful:
BTW, ctrlaltca, I am a member of the PRADO group under the hidden name javalizard.   While I could log in under my old user name and do the code changes myself, I thought my own branch may be better given my soon to be proposed changes/additions for PRADO. 

There were a number of serious flaws I found in my architectural use of PRADO and that is the source of these recent issues and changes.

The hope is to have these changes included in the 4.2,0 release.  Soon, I will look at #339 and make my changes to the system for enabling auto global listen.  It will be generally turned on, and some subclasses may have them turned off, like Lists, Maps, etc, collections.  It will be a general review of the __construct method esp. as relates to auto global listen.

belisoful:
It has taken until PHP 7.4 to really get Class Behaviors technically proper with WeakReferences.  A memory leak that big could be a problem for systems, which is why it has been off.   but with PHP 7.4, auto global listen can be turned on without the memory leaks from the lack of weakReferences.

framework/Web/Services/TPageService copy.php
```diff
@@ -0,0 +1,573 @@
+<?php
```
ctrlaltca, line 1: This file has probably been pushed by mistake

framework/Web/UI/WebControls/TPageLoadTime.php
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * TPageLoadTime class.
+ *
+ * Writes the amount of time taken from application start to rendering the contents of this control.
+ *
+ * @author Brad Anderson <belisoful@icloud.com>
+ * @package Prado\Web\UI\WebControls
+ * @since 4.2.0
+ */
+class TPageLoadTime extends TLabel
```
ctrlaltca, line 22: Do we really need a web control for this? TLogger already does the same, and its output can be routed to different places as needed

ctrlaltca:
Please, if you are implementing different changes, use different branches and create multiple PRs.
Eg. the TPageLoadTime changes are unrelated to the rest, they can be made in a different branch

framework/Web/Services/TPageService copy.php
belisoful: Indeed I was!  I don't know how this got in there.  I specifically did not check this file

framework/Web/UI/WebControls/TPageLoadTime.php
belisoful: There are webpages that link in the load time.  It was to make use of the application start time.

belisoful:
If you believe the startTime and initial memory usage are not good, I'll reverse that for you on this branch.  Some websites like to post the page load time at the bottom (or somewhere) of the page.  I don't know why they do that, but the TPageLoadTime is a use case for the application capturing its start time.

If I remember correctly, most platforms record their start time.

belisoful:
I am working on unifying the __construct methods' parent chain throughout the framework.  I am hoping this __construct change gets into the 4.2.0 release.  I am verifying my work today and running tests after.

belisoful:
I don't know what GitHub was doing. I only had one file selected for checkin.  It started checking in all the parent::__construct changes too!  I didn't mean for it to do that.  That is supposed to be a single large checkin.  I feel bad now.  making myself look back not using these tools right.  sigh.  ok.  back to work.

belisoful:
This set of updates is complete.  Have a look across the changes for the pull request, if you don't mind.

belisoful:
I'm working on getting these checks to complete.  All of these changes should be pulled.  They make the framework much more competitive and adaptable.  The CLI Application commands are important.  I'm working on a few other things.

belisoful:
not sure what the braces or binary_operator_spaces errors are coming from.

ctrlaltca:
This is becoming a really big change :) I'm not sure how to deal with it.
I'd like to merge at least the __construct changes soon, but i don't want to ruin your branch by cherry picking.
Do you mind if I ask you to move all the __construct changes in another new PR so that i can review, test and merge them?
We can then proceed to review and merge all the other changes the same way, one functional piece at a time.

belisoful:
these are changes I've been hanging onto for a while and should be folded back.  I've been testing the crap out of these things.  There are more changes forth-coming too, and also, hopefully, more discussion.  This new functionality should be a winner for the 4.2 release.  The changed file list is large due to the __construct update.  Dealing with that seems like a good first step.

It is a lot of functionality very quickly.  Sorry about that.  It's just, I've been sitting on this gold mine for a while and it should be shared.

I'll make a new PR with select changes.  I'll get back with you shortly with the new PR.

I have an account as a PRADO developer under the name Javalizard.  I just don't use that any more. Would it be more useful to be doing the merging myself in a controlled way?  like transfer over the changes by selected commits based on functionality?  If you need proof that I am Javalizard, I can log in to GitHub under that name and set a few of the issues to bugs.  We have some options.  just saying.

belisoful:
We'll talk on the other pull.  This pull is closed.

Issue #740, opened 2021-04-12
in Yii, behaviors can also be arrays with "class" and its properties as the other key=>values. PRADO should have it too


belisoful:
attaching behaviors has 2 types supported: string of behavior class name, and behavior instance, 

a 3rd type should be supported and is found in Yii: 
`['class'=>'MyObjectBehavior' property1="value1" prop2="value2" attribute1="value3"...]`

This would need to be supported in ClassBehaviors too.

I have the changes to TComponent and it works in unittest but my computer crashed and wasn't able to write the new unit tests yet.

ctrlaltca:
Fixed in #749

Issue #728, opened 2020-12-28
TComponent - PHP 7.4 Weak References required for storing instances of classes for Class  Behaviors


javalizard:
This has been a LONG time coming with the release of 7.4...  The TComponent class needs to be operating on Weak References for Class-wide behaviours when applying class behaviours to already instanced objects.

This should be easy to store weak references (rather than the objects per class) but it should only work for PHP 7.4, and be "backward compatible" where compatible.

This works without the weak references but would be better object management with weak references.  This is the perfect use case for weak references.

belisoful:
**Intention**
Turn getAutoGlobalListen to TRUE for a more generous ecosystem of code behavior and functionality.  There are branches that should have this variable turned on by default like TWebControl due to its potential usefulness. In PHP version 7.4, objects can be deleted without circular references in the global event handler (required for autoGlobalListen=true).  This memory leak is stopped before PHP 7.4 with autoGloballisten=false.

**Statement of Problem**
When getAutoGlobalListen is set to true, each TComponent is listening for a global event fxAttachClassBehavior.   Capturing this event adds a reference to the component in the static variable $_ue.  The issue is that the unlisten method is called in the __destruct so as to automate the tear down process.  The unlisten removes the reference, however the object won't destruct unless the reference is removed or not there.   catch-22, leading to a memory leak when relying on the automatic unlisten feature.

All global events, and events in general [both fx and on events, but not dy events], should be weak references.

Weak references will enables objects to be destructed and have the unlisten method called automatically.

Applications can add handlers to events in many ways, so the prior used formats must be maintained.

This adds overhead to instancing every object, but with the speed increases of PHP 7, this should be marginal in performance loss.  This can be tested.

This code change must be tested in versions of PHP before 7.4 for backward compatibility without weak references. TComponent regression is required in all PHP environments tested.

Solution:
adding the pseudo-code to attaching and detaching behaviors
    if handler is array and weak reference is available then      
             set the 0 item to the 0 item object wrapped in a WeakReference

Then in raise event, check if the 0 item object is a WeakReference and if it is, get the reference.  if null, continue.

belisoful:
The new attachEventHandler
public function attachEventHandler($name, $handler, $priority = null)
	{
		if(is_array($handler) && is_object($handler[0]) && class_exists('WeakReference') && !is_a($handler[0], 'WeakReference')) {
			echo("weak reference");
			$handler[0] = WeakReference::create($handler[0]);
		}
[...]

PR #749, opened 2021-04-22
TComponent updated for TWeakCallableCollection and Behaviors can instance from array definitions

belisoful:
#728 TComponent uses TWeakCallableCollection to store event handlers, that way garbage collection can delete objects that still have active handlers.  Because those handlers are weak references, the object won't get stuck in garbage collection heck and refuse to GC.  TWeakCallableCollection is to prevent memory leaks due to event handlers causing circular and hard references.

#740 Yii has the ability to instance classes from a class name string, using a class instance, and from an array defining the class and additional properties.

These additional properties are important to future features to be added.

### DIFF
--- a/framework/Caching/TDirectoryCacheDependency.php
+++ b/framework/Caching/TDirectoryCacheDependency.php
@@ -46,6 +46,7 @@ class TDirectoryCacheDependency extends TCacheDependency
 	public function __construct($directory)
 	{
 		$this->setDirectory($directory);
+		parent::__construct();
 	}
 
 	/**
--- a/framework/Caching/TFileCacheDependency.php
+++ b/framework/Caching/TFileCacheDependency.php
@@ -34,6 +34,7 @@ class TFileCacheDependency extends TCacheDependency
 	public function __construct($fileName)
 	{
 		$this->setFileName($fileName);
+		parent::__construct();
 	}
 
 	/**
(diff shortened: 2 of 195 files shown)


belisoful:
The Unit Tests cover all changes made.  TComponent using TWeakCallableCollection tests both extensively, and the TComponentTest checks that behaviors are instancing correctly in all instances.

I hope this is easy to merge.  I thought this was going to be harder to revert things and go through it incrementally.

Also, I'm thinking of doing the new events after this. like TAuthManager::onLogon, TApplication::onInitComplete, and TPageService::onPrePageRun. Then the autogloballisten turned on for ApplicationComponents and some TComponent subclasses.

belisoful:
The TComponentTest is changed to account for the error on the TComponent sub properties as well.  Isolating that one specific change in TComponentTest was difficult and is rolled into this check-in.   This new TComponent unit test will fail with the TComponent sub properties bug from before.

ctrlaltca (approved):
Code looks fine, now i'm up for some testing, especially with old php versions.
Even if I don't see a real cause why it should lead to any problem, better stay on the safe side

belisoful:
Indeed.  safety first.

FYI, I have tested this on 7.3.11 and 7.4 (without and with, respectively, WeakReference), and both pass unit test.  The fallback behavior works efficiently.

belisoful:
CTRL-ALT-C-A you are as much a code historian and librarian as much as an architect.

ctrlaltca:
No regression found, I'd say this can be safely merged.

Issue #738, opened 2021-04-07
TAuthManager needs onLogin, onLoginFailed, and onLogout events..


belisoful:
These are very significant events for a system and they should have actual events linked to their occurrences.  Other platforms allow hooks into this event, Wordpress, Drupal.

ctrlaltca:
Fixed in #751

Issue #735, opened 2021-04-02
General Inability of Modules to access page events.  No event between creation of page and running of page


belisoful:
If a module wants to handle a page event, there is no general event/method/location where a module can access a page events before the page events are run.  This is a severe bug as it limits functionality of all modules to be able to patch into page events.

Solution: In TPageService add a new event onPreRunPage that is executed in runPage between the Properties being set and the page being run.

`
	/**
	 * Executes a page.
	 * @param TPage $page the page instance to be run
	 * @param array $properties list of initial page properties
	 */
	protected function runPage($page, $properties)
	{
		foreach ($properties as $name => $value) {
			$page->setSubProperty($name, $value);
		}
		$this->onPreRunPage($page);
		$page->run($this->getResponse()->createHtmlWriter());
	}
	
	/**
	 * This is run just before the page is run.  Any part of the system can now patch into the page.
	 *	@param mixed what is passed as the parameter to the event
	 */
	public function onPreRunPage($param)
	{
		$this->raiseEvent('onPreRunPage',$this,$param);
	}
`

ctrlaltca:
This sounds like a nice addition to me

ctrlaltca:
Fixed in #751

PR #751, opened 2021-04-23
Added events to TAuthManager (onLogin, onLoginFailed, onLogout), TApplication::onInitComplete, TModule::dyPreInit, and TPageService::onPreRunPage:

belisoful:
#738 The TAuthManager::onLogin, onLoginFailed, onLogout events have been sorely needed for a long while, all the other platforms have triggers/handlers on these events.

#488 TApplication:onInitComplete resolves module initialization issues, especially in CLI applications.

#735 Enables TModules to access the TPage events before being raised with TPageService::onPreRunPage

The TModule::dyPreInit is raised by TApplication::internalLoadModule after the module is loaded but before the module initialization loop.  This is mainly for developers adding modules within modules, like say a UserManager Module within a WordPress Integration Module can add child modules to the loading modules for processing, and then initialize the children when the main module is initialized.

### DIFF
--- a/framework/Caching/TDirectoryCacheDependency.php
+++ b/framework/Caching/TDirectoryCacheDependency.php
@@ -46,6 +46,7 @@ class TDirectoryCacheDependency extends TCacheDependency
 	public function __construct($directory)
 	{
 		$this->setDirectory($directory);
+		parent::__construct();
 	}
 
 	/**
--- a/framework/Caching/TFileCacheDependency.php
+++ b/framework/Caching/TFileCacheDependency.php
@@ -34,6 +34,7 @@ class TFileCacheDependency extends TCacheDependency
 	public function __construct($fileName)
 	{
 		$this->setFileName($fileName);
+		parent::__construct();
 	}
 
 	/**
--- a/framework/Caching/TGlobalStateCacheDependency.php
+++ b/framework/Caching/TGlobalStateCacheDependency.php
@@ -36,6 +36,7 @@ class TGlobalStateCacheDependency extends TCacheDependency
 	public function __construct($name)
 	{
 		$this->setStateName($name);
+		parent::__construct();
 	}
 
 	/**
--- a/framework/Collections/TDummyDataSource.php
+++ b/framework/Collections/TDummyDataSource.php
@@ -36,6 +36,7 @@ class TDummyDataSource extends \Prado\TComponent implements \IteratorAggregate, 
 	public function __construct($count)
 	{
 		$this->_count = $count;
+		parent::__construct();
 	}
 
 	/**
--- a/framework/Collections/TList.php
+++ b/framework/Collections/TList.php
@@ -70,6 +70,7 @@ class TList extends \Prado\TComponent implements \IteratorAggregate, \ArrayAcces
 			$this->copyFrom($data);
 		}
 		$this->setReadOnly($readOnly);
+		parent::__construct();
 	}
 
 	/**
--- a/framework/Collections/TMap.php
+++ b/framework/Collections/TMap.php
@@ -39,7 +39,7 @@ use Traversable;
 class TMap extends \Prado\TComponent implements \IteratorAggregate, \ArrayAccess, \Countable
 {
 	/**
-	 * @var array internal data storage
+	 * @var array<mixed, mixed> internal data storage
 	 */
 	protected $_d = [];
 	/**
@@ -126,7 +126,7 @@ class TMap extends \Prado\TComponent implements \IteratorAggregate, \ArrayAccess
 	}
 
 	/**
-	 * @return array the key list
+	 * @return array<mixed> the key list
 	 */
 	public function getKeys()
 	{
@@ -201,7 +201,7 @@ class TMap extends \Prado\TComponent implements \IteratorAggregate, \ArrayAccess
 	}
 
 	/**
-	 * @return array the list of items in array
+	 * @return array<mixed, mixed> the list of items in array
 	 */
 	public function toArray()
 	{
@@ -259,7 +259,7 @@ class TMap extends \Prado\TComponent implements \IteratorAggregate, \ArrayAccess
 	/**
 	 * Returns the element at the specified offset.
 	 * This method is required by the interface \ArrayAccess.
-	 * @param int $offset the offset to retrieve element.
+	 * @param mixed $offset the offset to retrieve element.
 	 * @return mixed the element at the offset, null if no element is found at the offset
 	 */
 	public function offsetGet($offset)
@@ -270,7 +270,7 @@ class TMap extends \Prado\TComponent implements \IteratorAggregate, \ArrayAccess
 	/**
 	 * Sets the element at the specified offset.
 	 * This method is required by the interface \ArrayAccess.
-	 * @param int $offset the offset to set element
+	 * @param mixed $offset the offset to set element
 	 * @param mixed $item the element value
 	 */
 	public function offsetSet($offset, $item)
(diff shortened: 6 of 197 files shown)


ctrlaltca:
The first 3 changes are good to go.
The dyPreInit change looks a bit strange.
I understand that we need a way for code to patch into a module before it gets initialized. 
Having that single dynamic event triggered alone is a sort of "hidden gem" that's hard to find.
I would say "let's create a few TModule events like preInit(), postInit(), etc.." so that they can be easily documented, but that won't work since nothing can bind to that event before the module actually exists.
I then would say "let's add a new event to TApplication, eg. onModuleLoaded()", that could work and it would be easy to document aswell.
What do you think?

belisoful:
The entire dynamic event space 'dy' is raisable like a method. I think of them as signals to objects and its behaviors.  This would be a perfect use case for a dynamic event (dyPreInit). dynamic events are real time behavior enabled raised 'on' events within an entire 'dy' space.  dynamic events are designed to be light weight and used this exactly in this way. It's a new idea for inter-object communications that other frameworks do not have.  We should take advantage of it.  objects can implement the event/method or not. behaviors can implement the event or not.

It technically could be done either way

onModuleLoaded or onModulesLoaded could be populated by module __construct methods.  this seems like more changes than that of using dynamic events like this.  There are other changes that make use of dynamic events, like in TMap and TPriorityMap for processing not found items, adding and removing items.

The documentation issue could be resolved by having a discrete paragraph in each method php doc that calls dynamic events or for the class calling dynamic events.  there are other changes that make use of dynamic events like this.  I hope those changes don't spook people.

ctrlaltca:
I was proposing something simple like:

```php
	protected function internalLoadModule($id, $force = false)
	{
		// existing code

		$this->onModuleLoaded([$module, $configElement]);

		return [$module, $configElement];
	}

	/**
	 * Raises OnModuleLoaded event.
	 * This event is raised every time a new module is loaded, but before its 
	 * initialization method {@link TModule::init() } is called.
	 * Modules can use this event to patch into the module and inject dependencies,
	 * bind custom event hanlders or add behaviors to the module.
	 * @param array<TModule, mixed> $modinfo the just-loaded module object and its configuration
	 */
	public function onModuleLoaded($modinfo)
	{
		$this->raiseEvent('OnModuleLoaded', $this, $modinfo);
	}
```

Could it work for you?

belisoful:
Informing the whole application by 'on' event of one module Pre-loading, for each module, seems like overkill when the module just needs to be informed of being loaded.  Raising an application level event for an internal communication seems like overkill.

I like your solution but, having written the entire dynamic event code myself, I think such features and functionality should be used at some point without trepidation.  The 'dy' event space was created to compete with Drupal and Wordpress actions and filters, as a "real time" "on"-event intra-object communication feature.  Behaviors can hook events, handle dynamic events, and automatically register 'fx' events.  regular objects are designed to be failure resistant when implementing dynamic events.

the 'dy' event space is supposed to work like this.  this fits the use cases of dynamic events.  'dy' events are flyweight events, defining the event and using it in the same place.  PRADO has had dynamic events since their introduction in 3.2.3.  I think of all "on" methods as defining the limited events.  in "dy' events, the entire event space is defined and handled, any dynamic event can be called and it will return without error.  It is unique in that way.  **all objects have all dynamic events, always.   It is an inversion of limited thinking.  No other platform has dynamic events like this.  It is our advantage.**

_What exactly seems offensive to your sensibilities?  Is it documentation?  Is it using dynamic events themselves?_  I'm sure we can figure out a way to document dynamic events in a standard way.  I was hoping that several years would help people familiarize themselves with dynamic events and their usage.  I haven't seen much adoption and use of dynamic events... maybe because such inversion of thinking offends sensibilities, as stated in your message "looks a bit strange to me".

belisoful:
'on' events are mainly for communications to external objects.  'dy' events are for internal communications.  'fx' are for global communications.

I am suggesting that dyPreInit is an internal communication signal and not an external object communication signal.

belisoful:
A method in TModule for preInit($config) that gets overridden by subclasses, called in the same place, would work.  dyPreInit does the same thing.  I like the TModule::preInit($config) (overridden by subclasses when needed) method idea the best as a replacement for dyPreInit to address your concerns so far.

I didn't want to have to add a default method for preInit to TModule.  As someone of the eastern philosophy, the less the better.

ctrlaltca:
> _What exactly seems offensive to your sensibilities? Is it documentation? Is it using dynamic events themselves?_

I'm not against dynamic events at all, but I'm not comfortable in introducing only one or a few of them in seemly random places inside the core of PRADO, it looks inconsistent to me. I'd prefer to create a plan on what dynamic method will be implemented in the core, and a solid rationale on where they are available and why.

I'm sorry, maybe it's me non understanding the full vision of what you are trying to achieve.
A few random questions that could help me making my mind clear on this topic:
 * as an app developer using PRADO, how am i supposed to know that dyPreInit() exists on modules? Ok, we could document it, but being the only one dynamic method in core it would be quite a hidden gem to find;
 * I would suppose that since dyPreInit() exists, maybe it has a dyPostInit() counterpart, or others? If not, what's the rationale behind this? I get dyPreInit is the only one that is needed by now to implement your next ideas on the code, but including in core a functionality that is used only once in the code smells added complexity to me. 
 * What other dynamic methods are implemented in the core? None actually, sorry if i repeat myself but I'd like to understand if there's a plan on what others dynamic events will get implemented before start adding them, to ensure coherency.
 * Since dyPreInit is called on modules being loaded, why it's not called on other classes/objects/whatever? How am i supposed to know what dynamic events are raised in each class?

ctrlaltca:
Sorry, re-reading this: 

> Informing the whole application by 'on' event of one module Pre-loading, for each module, seems like overkill when the module just needs to be informed of being loaded. Raising an application level event for an internal communication seems like overkill.

So you basically need __construct()?

> A method in TModule for preInit($config) that gets overridden by subclasses, called in the same place, would work

Why not override init($config) in subclasses then? Sorry I still don't get it :(

belisoful:
I'm likely not describing it properly... my apologies.

preInit is needed for after the module properties are loaded but before init is called.  if a module loads its own internal modules that are externally referenced, Init is not going to be enough.  The sub-modules should be loaded prior to init being called otherwise external modules that rely on sub-modules being loaded may not be able to reference the internal sub-modules.

Use case: WPIntegrator has user manager, auth manager, and parameter loader sub-modules that are loaded during preInit and then initialized with the normal module init.  The sub-modules ids are set from the module properties, so the properties must be set.  If the sub-modules are loaded and initialized in the module init, then external modules linking to the user and auth sub-modules cannot find them as they may not be  loaded with the normal Application loader under module init.

How about I add TModule::preInit method?

belisoful:
I am coding an experimental preInit into TModule... Here is the issue, should we make iModule require preInit or not? Because technically, if we make a method in TModule::preInit then iModule will require it and break backward compatibility.  ALL IModules would need preInit.  That's a big requirement for users to update their code to upgrade like that.

dyPreInit doesn't require any changes to IModule, I don't want to change IModule to call preInit in TApplication as a required method, even if the TModule implements it for most modules.  Your thoughts?

using 'dy' event space as optional method space, exactly like this, is a valid use of 'dy' events.

I was going to test out preInit in TModules until IModules issues during coding came up.  Now I'm solidly in preference of dyPreInit as a solution for backward compatibility, reducing requirements, and simplicity.

belisoful:
dynamic events need to be implemented through PRADO. TApplication could probably use more dynamic events, like TComponent.

A good example is in TTheme on my branch of how I propose using dynamic events.  
See this https://github.com/belisoful/prado/blob/5dd8189c15af7b0c4d4adadb8116f11d121ab45a/framework/Web/UI/TTheme.php#L88

ctrlaltca:
Well thank you, I totally missed the point before and now i have a clearer idea on what you are trying to achieve.

PR #753, opened 2021-04-24
Merging AutoGlobalListen as true for TApplicationComponent and TUser

belisoful:
This enables class behaviors by default on __construct for all TApplicationComponents and TUser.  

Are there any other TComponent subclasses that should be class behavior enabled by default?  TWeakCallableCollection cannot have AutoGlobalListen because that would cause recursion, and that is accounted for and already pre-set just in case someone is messing around and sets TComponent Autogloballisten to true, TWeakCallableCollection is protected.

This should be tested on PHP 7.4, with WeakReference, and on PHP 7.3- without WeakReference.

### DIFF
--- a/framework/Caching/TDirectoryCacheDependency.php
+++ b/framework/Caching/TDirectoryCacheDependency.php
@@ -46,6 +46,7 @@ class TDirectoryCacheDependency extends TCacheDependency
 	public function __construct($directory)
 	{
 		$this->setDirectory($directory);
+		parent::__construct();
 	}
 
 	/**
--- a/framework/Caching/TFileCacheDependency.php
+++ b/framework/Caching/TFileCacheDependency.php
@@ -34,6 +34,7 @@ class TFileCacheDependency extends TCacheDependency
 	public function __construct($fileName)
 	{
 		$this->setFileName($fileName);
+		parent::__construct();
 	}
 
 	/**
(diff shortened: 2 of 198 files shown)


ctrlaltca:
From a test point of view this works fine.
I'm now profiling it, but the results are quite unsatisfactory.
Tests on prado-demos/helloworld: before 90ms, after 200ms
Tests on a huge page from an internal app: before 1.4 sec, after 3.0 sec.

The biggest offender seems to be the array_filter() calls inside listen() and unlisten() used to determine the fx methods.
They are run for every component in the page, and takes up an average of 52 - 55% of the total page render time.

ctrlaltca:
Before:
before

After:
after

ctrlaltca:
With this performance loss I'm not willing to keep this enabled for every component.
The performance loss is not due to event handling itself, but to the single line:
```
$fx = array_filter(get_class_methods($this), [$this, 'filter_prado_fx']);
```
that gets executed both in listen() and in unlisten(). Caching that array in a variable would probably already halve the performance loss, but it's still very expensive to loop over all the defined class methods and strncasecmp() over them.
It would be nice to find a way where this performance loss is minimized only to the components where actually at least one fx is implemented, without having to loop over all the methods.

belisoful:
in my testing I found some performance loss but apparently your tools are MUCH better for determining these things.

I was not aware of just how expensive this function is.  Let me see what I can do to speed this up.

PR #754, opened 2021-04-24
TComponent Added caching to fx event processing.

belisoful:
Hopefully this will speed things up significantly to an acceptable place.  I don't have the speed test tools you have so I can't test this as extensively as you have done.

### DIFF
--- a/.github/workflows/prado.yml
+++ b/.github/workflows/prado.yml
@@ -44,9 +44,6 @@ jobs:
       - name: Validate code syntax using php-cs-fixer
         run: php-cs-fixer fix --dry-run --format=checkstyle | cs2pr
 
-      - name: Validate code using phpstan
-        run: phpstan analyse --error-format=checkstyle --level=1 framework | cs2pr
-
       - name: Get composer cache directory
         id: composer-cache
         run: echo "::set-output name=dir::$(composer config cache-files-dir)"
@@ -60,6 +57,9 @@ jobs:
 
       - name: Install Composer dependencies
         run: composer install --no-progress --prefer-dist --optimize-autoloader
+
+      - name: Validate code using phpstan
+        run: phpstan analyse --error-format=checkstyle --level=1 framework | cs2pr
 
       - name: Setup mysql service
         run: |
--- a/framework/Caching/TDirectoryCacheDependency.php
+++ b/framework/Caching/TDirectoryCacheDependency.php
@@ -46,6 +46,7 @@ class TDirectoryCacheDependency extends TCacheDependency
 	public function __construct($directory)
 	{
 		$this->setDirectory($directory);
+		parent::__construct();
 	}
 
 	/**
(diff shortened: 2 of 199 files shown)


belisoful:
I thought array_filter was going to be much faster and light weight than your testing tools shows.  Apparently, it is called enough to dent the performance.   :(   I really hope this helps and resolves the performance issue.

belisoful:
If this doesn't work to speed things up, how about changing over from AutoGlobalListen fxevents to a specific list of events that are handled, like Behaviors with 'on' events.
https://github.com/pradosoft/prado/blob/3d13abeb5eb909755b68b98d051d74295bb31e09/framework/Util/TBehavior.php#L31

providing an array of fx events to patch will stop the array_filter search for fx methods.

I still really like the automation of automagically attaching fx global events.

ctrlaltca:
First results:
 * in the helloworld demo it doesn't really make a big difference, i suppose that because there are only a few components.
 *  in the huge page tests (that originally ran in 1.4secs) we're down from 3 secs to 2.1 secs, so this patch cuts down 2/3 of the performance loss.
 
Right now the remaining performance loss is mostly due to the loop in listen/unlisten:

```php
foreach ($fx as $func) {
	$this->getEventHandlers($func)->add([$this, $func]);
}
```

Every class subclassing TComponent  has fxAttachClassBehavior and fxDetachClassBehavior.
My my huge page i get 3400 iterations on this loop, i guess i have about 850 application components * 2 fx methods (attach/detach) * 2 called methods (listen/ulisten).

belisoful:
hmmm...   I'm glad the caching had such a great result.

My thoughts: Modules mainly are the components attaching class behaviors.  TComponent AutoGlobalListen functionality is for applying class behaviors upon already instanced objects.  By the time the TPage is executed, and TControls constructed, class behaviors should technically all be installed.   TControls won't need AutoGlobalListen because new Class Behaviors would likely NOT be added in the middle of a page run, all the class behaviors would be added before the page is run, ideally.

**Solution: Turn off AutoGlobalListen for TControls.  This should limit execution of the offending lines.**

I'm looking into PHP profiling tools, like you have.  any suggestions?

ctrlaltca:
> I'm looking into PHP profiling tools, like you have. any suggestions?

I use the[Xdebug](http://xdebug.org/docs/profiler) php extension's profiler. It outputs files in cachegrind format that can be viewed with KCachegrind, WinCachegrind or webgrind.

ctrlaltca:
With the last version of this PR we're down from 2.1 to 1.65 secs, so now there's just a minimal overhead.
I made some tests just removing from TComponent all the code related to dy, fx and behaviors and the page ran in 1.1 sec.
So there's still some room for further improvement, but this looks good enough for me right now.

ctrlaltca:
I'd commit this as it is by now, is it ok for you?

belisoful:
Indeed, I believe this change is complete.  How are the latest changes looking?

Issue #744, opened 2021-04-20
Converting standard LTR into RTL... is there a standardized mechanism to determine RTL?


belisoful:
Is there a method for determining the text direction LTR or RTL within globalization? 

Here are the languages that should be RTL:
from https://lingohub.com/academy/best-practices/rtl-language-list
`
['ar', 'arc', 'dv', 'fa', 'ha', 'he', 'khw', 'ks', 'ku', 'ps', 'ur', 'yi']`

This could be hard coded or the translator could be used to translate the text 'ltr' and check the output for 'rtl'.  This is how wordpress does RTL determination.

ctrlaltca:
To determine id RTL or LTR:
```php
$o = new \Prado\I18N\core\CultureInfo('ar');
return $o->findInfo('layout/characters');
```

belisoful:
Sir, you are an impressive person!

This seems like something that Globalization should encapsulate.  RTL is an important application variable for globalization.

This outputs the text (without quotes):
`'right-to-left'`

ctrlaltca:
When I rewrote the globalization classes to use php's intl extension i implemented the bare minimum necessary to get all the existing classes working. There's a lot of room for improvements

belisoful:
Ah.  Thank you for the re-write!  I cannot overstate how brilliant and hard working you are.

The RTL determination may be something I'll add to TGlobalization shortly.  If that is the case, I'll probably add *rtl.css support in themes. That way when globalization is RTL, the RTL extension css files are automatically added in the theme, otherwise *rtl.css will not be added.

Wordpress has RTL determination and automatically adds RTL css files when needed.  This is something PRADO should be doing as well.

belisoful:
https://github.com/pradosoft/prado/pull/756/files

This pull request adds Globalization RTL determination and Ttheme RTL support.

belisoful:
RTL support added

PR #756, opened 2021-04-26
General Utility Update and Addition

belisoful:
This adds:
-TException support for multiple error files
-TGlobalization checking for RTL
-TApplication Adds an event dyInit during Parameter loading of classes, to make them a bit more useful.
-TTheme adds RTL support, removing 'rtl" css files, then if RTL is true, adding the rtl css at the end.
-TStyle Adds Border Radius
-TPanelStyle  Adds Box Shadow
-TTemplateControl Adds support for nested TTemplateControl MasterClass.  eg MasterLayout has placeholder A, B, and C,  CommonLayout uses MasterLayout as its MasterClass and defines content on A and C and not B. Then the Page defines B and uses CommonLayout for the Page MasterClass... This would throw an error, now the content is injected up the chain.
(This is useful if, say Admin pages have a separate MasterClass [CommonLayout vs AdminLayout] for header and footer but a unified MasterLayout)

### DIFF
--- a/framework/Exceptions/TException.php
+++ b/framework/Exceptions/TException.php
@@ -64,6 +64,7 @@ class TException extends \Exception
 	/**
 	 * Adds to the various files to read when rendering an error
 	 * @param string $file the extra message file
+	 * @since 4.2.0
 	 */
 	public static function addMessageFile($file)
 	{
--- a/framework/Exceptions/TException.php
+++ b/framework/Exceptions/TException.php
@@ -97,9 +97,9 @@ class TException extends \Exception
 						[$code, $message] = array_merge(explode('=', $entry, 2), ['']);
 						self::$_messageCache[$msgFile][trim($code)] = trim($message);
 					}
-					$value = self::$_messageCache[$msgFile][$key] ?? $value;
 				}
 			}
+			$value = self::$_messageCache[$msgFile][$key] ?? $value;
 		}
 		return $value;
 	}
--- a/framework/Exceptions/TException.php
+++ b/framework/Exceptions/TException.php
@@ -37,6 +37,7 @@ use Prado\TPropertyValue;
 class TException extends \Exception
 {
 	private $_errorCode = '';
+	private static $_messagefiles = [];
 	protected static $_messageCache = [];
 
 	/**
@@ -59,6 +60,22 @@ class TException extends \Exception
 		}
 		parent::__construct(strtr($errorMessage, $tokens));
 	}
+	
+	/**
+	 * Adds to the various files to read when rendering an error
+	 * @param string $file the extra message file
+	 */
+	public static function addMessageFile($file)
+	{
+		if (preg_match('/^(.*)(-.{2.4})?\.(.{2,4})$/', $file, $matching)) {
+			$lang = Prado::getPreferredLanguage();
+			$msgFile = $matching[1] . '-' . $lang . '.' . $matching[2];
+			if (is_file($msgFile)) {
+				$file = $msgFile;
+			}
+		}
+		TException::$_messagefiles[] = $file;
+	}
 
 	/**
 	 * Translates an error code into an error message.
@@ -67,18 +84,23 @@ class TException extends \Exception
 	 */
 	protected function translateErrorMessage($key)
 	{
-		$msgFile = $this->getErrorMessageFile();
+		$msgFiles = TException::$_messagefiles;
+		$msgFiles[] = $this->getErrorMessageFile();
+		$value = $key;
 
 		// Cache messages
-		if (!isset(self::$_messageCache[$msgFile])) {
-			if (($entries = @file($msgFile)) !== false) {
-				foreach ($entries as $entry) {
-					[$code, $message] = array_merge(explode('=', $entry, 2), ['']);
-					self::$_messageCache[$msgFile][trim($code)] = trim($message);
+		foreach ($msgFiles as $msgFile) {
+			if (!isset(self::$_messageCache[$msgFile])) {
+				if (($entries = @file($msgFile)) !== false) {
+					foreach ($entries as $entry) {
+						[$code, $message] = array_merge(explode('=', $entry, 2), ['']);
+						self::$_messageCache[$msgFile][trim($code)] = trim($message);
+					}
+					$value = self::$_messageCache[$msgFile][$key] ?? $value;
 				}
 			}
 		}
-		return self::$_messageCache[$msgFile][$key] ?? $key;
+		return $value;
 	}
 
 	/**
--- a/framework/TApplication.php
+++ b/framework/TApplication.php
@@ -1119,6 +1119,7 @@ class TApplication extends \Prado\TComponent
 	 * request.  This call is important for CLI/Shell applications that do not have
 	 * a web service lifecycle stack.  This is the first and last event for finalization
 	 * of any loaded modules in CLI/Shell mode.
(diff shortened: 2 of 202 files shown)


belisoful:
https://github.com/pradosoft/prado/issues/744

framework/Web/UI/WebControls/TStyle.php
```diff
@@ -174,6 +174,32 @@ public function setBorderWidth($value)
...
+	 * @since 4.2.0
+	 */
+	public function setBorderRadius($value)
+	{
+		if (trim($value) === '') {
+			unset($this->_fields['-moz-border-radius']);
+			unset($this->_fields['-webkit-border-radius']);
+			unset($this->_fields['border-radius']);
+		} else {
+			$this->_fields['-moz-border-radius'] = $value;
```
ctrlaltca, line 197: Do we really need -moz-border-radius and -webkit-border-radius?
The generic border-radius is well supported even by ancient browsers: https://caniuse.com/?search=border-radius
-moz-border-radius only add compatiblity for firefox 2.x
-webkit-border radius for ancient chrome versions

framework/Web/UI/WebControls/TPanelStyle.php
```diff
@@ -103,6 +110,11 @@ public function addAttributesToRender($writer)
 				$this->setStyleField('direction', 'rtl');
 			}
 		}
+		if ($boxShadow = $this->getBoxShadow()) {
+			$this->setStyleField('-moz-box-shadow', $boxShadow);
+			$this->setStyleField('-webkit-box-shadow', $boxShadow);
+			$this->setStyleField('box-shadow', $boxShadow);
```
ctrlaltca, line 116: Please drop the vendor-specific -moz-box-shadow and -webkit-box-shadow, they don't really add support for any browser

framework/Web/UI/TTheme.php
```diff
@@ -331,4 +337,29 @@ protected function setJavaScriptFiles($value)
...
+				(isset($segs[2]) && (substr_compare(strtolower($segs[count($segs) - 3]), 'rtl', -3) === 0))) {
+				$rtlCss[] = $url;
+				unset($this->_cssFiles[$key]);
+			}
+		}
+		if ($globalization = $this->getApplication()->getGlobalization()) {
+			if ($globalization->getIsCultureRTL()) {
+				$this->_cssFiles = array_merge($this->_cssFiles, $rtlCss);
+			}
+		}
```
ctrlaltca, line 363: This is a bit overly complicated, removing things from an array to add them back later.
I would rework it (in pseudocode):
```
// rtl: no css needs to be removed
if($globalization = ... && $globalization->getIsCultureRTL())
  return;

// remove rtl css files
foreach($this->_cssFiles as $key => $url) {
	if($this->getIsRtlCssFile($url)) {
	  unset($this->_cssFiles[$key]);
	}
}

```

Also, moving the actual comparison to an external "getIsRtlCssFile()" method would clarify its meaning

framework/TApplication.php
```diff
@@ -979,6 +979,7 @@ public function applyConfiguration($config, $withinService = false)
 				foreach ($parameter[1] as $name => $value) {
 					$component->setSubProperty($name, $value);
 				}
+				$component->dyInit();
```
ctrlaltca, line 982: This is ok by now; For the future, it would be nice to investigate a more generic way to call these dynamic methods.

framework/Exceptions/TException.php
```diff
@@ -67,18 +85,23 @@ public function __construct($errorMessage)
...
+					foreach ($entries as $entry) {
+						[$code, $message] = array_merge(explode('=', $entry, 2), ['']);
+						self::$_messageCache[$msgFile][trim($code)] = trim($message);
+					}
+					$value = self::$_messageCache[$msgFile][$key] ?? $value;
 				}
 			}
 		}
-		return self::$_messageCache[$msgFile][$key] ?? $key;
+		return $value;
```
ctrlaltca, line 104: This looks like an error, if the $_messageCache is already populated $value is never set

ctrlaltca (changes requested):
Please have a look at the inline comments

framework/Web/UI/WebControls/TStyle.php
belisoful: I guess I'm that old school.  You got it.

framework/Web/UI/WebControls/TPanelStyle.php
belisoful: will do.

framework/Web/UI/TTheme.php
belisoful: My thinking on this that LTR should first have RTL css removed by default.  This way LTR are protected from RTL css enabled themes.  Automatically including RTL (when there is no globalization) would always make themes RTL by default if the theme supports RTL, that's not preferred behavior.
Then if there is globalization and it's RTL, then the CSS is added _at the end_ of all css.  The "rtl css at the end" aspect is important for CSS priority of RTL.  The files later would override the LTR css with redefinitions. Without being re-added at the end would mean the RTL css would be based on file name order rather than having RTL css after all LTR css.

I am ambivalent on a getIsRtlCssFile method.  If you think it'll be good, I'll get on it.

framework/TApplication.php
belisoful: I'll check in the TTheme behavior (when this is resolved) enabling with dynamic events through it for how dynamic events should be used.  dynamic events have multiple use cases. I believe this is one of them.  I'm open to rethinking these things.  I'm not sure I understand making the dynamic event method space more generic.

framework/Exceptions/TException.php
belisoful: Line 90 has $value = $key;
and line 100 has $value = self::$_messageCache[$msgFile][$key] ?? $value;

Looping through the possible multiple files is an important aspect of the new method

framework/Web/UI/TTheme.php
ctrlaltca: You have a good point on this, i missed that.
You can keep it as it is.

framework/Exceptions/TException.php
ctrlaltca: Line 100 get executed only if `!isset(self::$_messageCache[$msgFile])`
So if the file has already been cached before, it's never executed (or maybe I'm overlooking this?)

framework/Web/UI/WebControls/TStyle.php
belisoful: changed

framework/Web/UI/WebControls/TPanelStyle.php
belisoful: changed

framework/Exceptions/TException.php
belisoful: my thinking may be off with this... let me see...

belisoful: I'm glad we are going over these.  where ever my brain was for that error is beyond me.  lol.  I have the correction here.

belisoful: corrected

framework/TApplication.php
belisoful: I see this as a valid use case for intra-object communication via 'dy' events.  It may look like a method but it's an event.

ctrlaltca (approved):

PR #757, opened 2021-04-27
TTheme added support for behaviors via dynamic events

belisoful:
If the additions for TApplication to preinit and init parameter classes raised your alarms, this will surely be triggering.  However, the point of this pull request is to discuss various usages of dynamic events.

Wordpress has do_action and action_filter (or something like that, if I recall off the top of my head).  Other platforms can add "events" on any keyword without defining the event.  

ON events are for one to many (external)
DY events are for one to one and one to many (internal, to behaviors on the object). External objects don't need to know the internal workings of an object, but the behavior does.

Yii uses ON events more extensively, intra-object.  PRADO has DY events and is more flexible and real time than ON events.

What is funny/odd about using dynamic events to you?

### DIFF
--- a/.github/workflows/prado.yml
+++ b/.github/workflows/prado.yml
@@ -36,13 +36,16 @@ jobs:
         with:
           php-version: ${{ matrix.php-versions }}
           extensions: ctype, dom, intl, json, mbstring, memcached, pdo_mysql, pdo_pgsql, openssl, pcre, spl, zlib
-          tools: php-cs-fixer, cs2pr
+          tools: php-cs-fixer, phpstan, cs2pr
 
       - name: Validate composer.json and composer.lock
         run: composer validate --strict
 
-      - name: Validate code syntax
+      - name: Validate code syntax using php-cs-fixer
         run: php-cs-fixer fix --dry-run --format=checkstyle | cs2pr
+
+      - name: Validate code using phpstan
+        run: phpstan analyse --error-format=checkstyle --level=1 framework | cs2pr
 
       - name: Get composer cache directory
         id: composer-cache
--- a/.github/workflows/prado.yml
+++ b/.github/workflows/prado.yml
@@ -44,9 +44,6 @@ jobs:
       - name: Validate code syntax using php-cs-fixer
         run: php-cs-fixer fix --dry-run --format=checkstyle | cs2pr
 
-      - name: Validate code using phpstan
-        run: phpstan analyse --error-format=checkstyle --level=1 framework | cs2pr
-
       - name: Get composer cache directory
         id: composer-cache
         run: echo "::set-output name=dir::$(composer config cache-files-dir)"
@@ -60,6 +57,9 @@ jobs:
 
       - name: Install Composer dependencies
         run: composer install --no-progress --prefer-dist --optimize-autoloader
+
+      - name: Validate code using phpstan
+        run: phpstan analyse --error-format=checkstyle --level=1 framework | cs2pr
 
       - name: Setup mysql service
         run: |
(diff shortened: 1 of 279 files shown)


belisoful:
Also consider that dynamic events are lighter weight than raiseEvent.  a DY event is less overhead than raiseEvent.

belisoful:
To assist in thinking about this, I have a question.  for myself as well.  How would you change the dynamic events in TComponent?  remove?  add more in various places? re-implement in another pattern?

Should TModule init have the dynamic event: 
`
public function init($config) {
dyInit($config);
}
`
?

belisoful:
dynamic events were the natural evolution of ON events meets behaviors.  Yii implements behaviors without this extension to dynamic events.  I believe it is a part of our strength and should be used, to that end, for making the framework as flexible as possible.

belisoful:
In typical object-method calls one call is one method.
`
...
$this->aMethod($param1);
...

----

public function aMethod($param){...}
` 

However, with behaviors, like Yii, there is a need for one event/call to then execute upon multiple of the same methods in multiple object endpoints internally.  I believe Yii's behavior model is superior to WordPress broadcast-receiver do_action/action_filter.   Yii doesn't make events actually look and program like method calls.

the model where events actually look, act, and simulate method calls is what makes the dynamic events so interesting.  'dy' events are an infinite 'on' event "space" of infinite flexibility between objects and their behaviors.  dynamic events operate in a JIT model rather than a predefined compiled model.

One of my use cases for 'fx' global events was for populating navigation menus from modules using something like a TEventContent webcontrol.  see this:

https://github.com/belisoful/prado/blob/master/framework/Web/UI/TEventContent.php

'fx' global events resolves the issue of unknown components talking to each other.  "Message to the aether" style.  Wordpress does that REALLY well, that is their only real model of "events."

ctrlaltca:
Don't worry, I'm not gonna let this stale here because I'm against the new event types or their use. I'm just a bit overwhelmed with my work, and i want to properly test and understand this.

> To assist in thinking about this, I have a question. for myself as well. How would you change the dynamic events in TComponent? remove? add more in various places? re-implement in another pattern?
> 
> Should TModule init have the dynamic event:
> `public function init($config) { $this->dyInit($config); }`
> ?

In my dreams it would be nice to have a set of dynamic events that gets raised by specific classes.
In your example with TModule, it would be nice to know that while working with every TModule subclass a dyInit() event will be raised; this can be properly documented in TModule itself so that people can confidently use it.

The same way, you are adding a few dynamic events to TTheme/TStyleSheet. Maybe others are needed/desirable?
I have no idea :)

belisoful:
No worries.  I know how life goes.  This PR poses questions about new dimensions of PRADO.  Great new flexibility (with performance) is my hope.

The unification of the __construct chain is extremely useful.  one of the coming pulls will be for the TBehaviorsModule that installs behaviors across the app.  Included with it will be a TParameterizeBehavior that allows any Property on any object to be set from parameter.  It has an interesting sub-behavior that attaches to the Application Parameter TMap and catch any changes and route it back to the setProperty method.   There are some dynamic events for TMap for No Item, Add Item, and Delete item that the sub-behavior relies upon.

https://github.com/belisoful/prado/blob/master/framework/Util/Behaviors/TParameterizeBehavior.php

So yes, there are a few other places that I've been actively finding for select new dynamic events.  e.g. a dynamic event dyNoItem in TMap allows for lazy loading of Parameters.  Which there is a behavior for that: TMapLazyLoadBehavior.

As you are looking into dynamic events and this PR, if/when you have the time, have a look at TMap with dynamic events and TMap*Behaviors.

https://github.com/belisoful/prado/blob/41ac0779a7406bf2bd143673e76a79dcff2f782d/framework/Collections/TMap.php#L147

and its associated TMapLazyLoadBehavior

https://github.com/belisoful/prado/blob/master/framework/Util/Behaviors/TMapLazyLoadBehavior.php


and 

https://github.com/belisoful/prado/blob/41ac0779a7406bf2bd143673e76a79dcff2f782d/framework/Collections/TMap.php#L161

with its associated TMapRouteBehavior

https://github.com/belisoful/prado/blob/master/framework/Util/Behaviors/TMapRouteBehavior.php

Issue #758, opened 2021-05-01
THttpSession throws "Cannot regenerate session id - headers already sent" warning when changing users in CLI with text printed on the screen


belisoful:
When operating in CLI mode, Switching users after text has been printed causes session_regenerate_id to throw a warning:

[Warning] session_regenerate_id(): Cannot regenerate session id - headers already sent (@line 190 in file /Applications/Ampps/www/app/protected/vendor/pradosoft/prado/framework/Web/THttpSession.php).   
#0 [internal function]: Prado\Prado::phpErrorHandler(2, 'session_regener...', '/Applications/A...', 190, Array)
#1 ${PradoFramework}/Web/THttpSession.php(190): session_regenerate_id(true)
#2 ${PradoFramework}/Security/TAuthManager.php(398): Prado\Web\THttpSession->regenerate(true)
#3 ${PradoFramework}/Security/TAuthManager.php(414): Prado\Security\TAuthManager->updateSessionUser(Object(Prado\Security\TUser))
#4 ${PradoFramework}/Util/Cron/TCronModule.php(275): Prado\Security\TAuthManager->switchUser('admin')

ctrlaltca:
Fixed in #759

PR #759, opened 2021-05-02
TModule::dyInit and subclass init chaining, TParameterModule calls dyInit on class parameters, and added method param for header scripts to load async

belisoful:
TClientScriptManager::registerHeadScriptFile takes async parameter

#758 THttpSession doesn't call session_regenerate_id when in CLI

Chaining of TModule::init and added dyInit($config) for TModule behaviors.

Corrected TParameterModule to call dyInit on class parameters (like in TApplication)

### DIFF
--- a/framework/TApplication.php
+++ b/framework/TApplication.php
@@ -1119,6 +1119,7 @@ class TApplication extends \Prado\TComponent
 	 * request.  This call is important for CLI/Shell applications that do not have
 	 * a web service lifecycle stack.  This is the first and last event for finalization
 	 * of any loaded modules in CLI/Shell mode.
+	 * @since 4.2.0
 	 */
 	public function onInitComplete()
 	{
--- a/framework/TApplication.php
+++ b/framework/TApplication.php
@@ -979,6 +979,7 @@ class TApplication extends \Prado\TComponent
 				foreach ($parameter[1] as $name => $value) {
 					$component->setSubProperty($name, $value);
 				}
+				$component->dyInit();
 				$this->_parameters->add($id, $component);
 			} else {
 				$this->_parameters->add($id, $parameter);
--- a/framework/TApplication.php
+++ b/framework/TApplication.php
@@ -58,6 +58,7 @@ use Prado\Util\TLogger;
  * TApplication maintains a lifecycle with the following stages:
  * - [construct] : construction of the application instance
  * - [initApplication] : load application configuration and instantiate modules and the requested service
+ * - onInitComplete : this event happens right after after module and service initialization. This event is particularly useful for CLI/Shell applications
  * - onBeginRequest : this event happens right after application initialization
  * - onAuthentication : this event happens when authentication is needed for the current request
  * - onAuthenticationComplete : this event happens right after the authentication is done for the current request
@@ -278,7 +279,8 @@ class TApplication extends \Prado\TComponent
 	 * Constructor.
 	 * Sets application base path and initializes the application singleton.
 	 * Application base path refers to the root directory storing application
-	 * data and code not directly accessible by Web users.
+	 * data and code not directly accessible by Web users.  This captures the
+	 * start time and initial memory usage.
 	 * By default, the base path is assumed to be the <b>protected</b>
 	 * directory under the directory containing the current running script.
 	 * @param string $basePath application base path or configuration file path.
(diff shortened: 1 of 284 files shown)


belisoful:
Most of this PR is straight forward.  This PR is looking for is a review of adding the async option to TClientScriptManager::registerHeadScriptFile.

The addition is backward compatible.  I just wanted another set of eyes on that change, briefly, before merging.

I was adding a google analytics script to the head and they want it to be async.  That's a very good idea fo external scripts like google's that are page view/state independent.

framework/Web/UI/TClientScriptManager.php
```diff
@@ -576,11 +576,12 @@ public function getStyleSheetCodes()
...
 	 * @param string $key a unique key identifying the file
 	 * @param string $url URL to the javascript file
+	 * @param bool $async load the javascript file asynchronously, default false
 	 */
-	public function registerHeadScriptFile($key, $url)
+	public function registerHeadScriptFile($key, $url, $async = false)
 	{
 		$this->checkIfNotInRender();
-		$this->_headScriptFiles[$key] = $url;
+		$this->_headScriptFiles[$key] = $async ? [$url, $async] : $url;
```
ctrlaltca, line 584: This probably breaks rendering of new script files during an ajax call, currently used by at least THtmlArea.
On callback, TActivePageAdapter::renderResponse() retrieves a list of the javascript files and renders their names inside a <X-PRADO-SCRIPTLIST> tag that gets checked by clientside javascript for missing scripts.
The clientside javascript expect plain urls.

framework/Web/THttpSession.php
```diff
@@ -187,7 +187,9 @@ public function destroy()
 	public function regenerate($deleteOld = false)
 	{
 		$old = $this->getSessionID();
-		session_regenerate_id($deleteOld);
+		if (php_sapi_name() !== 'cli') {
```
ctrlaltca, line 190: Why use http sessions at all on a CLI app? Is it because it shares the same application.xml as a web application?
This looks ok to avoid the error, but probably we could disengage the use of http sessions on CLI apps earlier.

ctrlaltca (changes requested):
The TModule::init() changes looks fine.
I have 2 comments on the other changes, please have a look.

framework/Web/THttpSession.php
belisoful: I'll have a look at removing httpsession from cli apps.   Great idea. I've been working on perfecting other aspects and this clearly didn't get my full attention.

belisoful: I wasn't sure how deep to look at this solution.

belisoful: Just one call seems to be what is needed.

framework/TApplication.php
```diff
@@ -1256,7 +1256,7 @@ public function flushOutput($continueBuffering = true)
 	public function onEndRequest()
 	{
 		$this->flushOutput(false); // flush all remaining content in the buffer
-		$this->saveGlobals();  // save global state
 		$this->raiseEvent('OnEndRequest', $this, null);
+		$this->saveGlobals();  // save global state
```
ctrlaltca, line 1260: Please note that the globalstate has already been saved on the onSaveState step.
This call to saveGlobals() is only a last-time fallback to ensure globals are saved, but changing the global sta ein the flush/OnEndRequest step will trigger a second forced saving with all the relative performance loss.

belisoful:
Regarding the final fallback saveGlobals.  I realize it is a fallback, if there are changes to the global state onEndRequest, those changes should be automatically captured.

Use case: I'm working on the Cron Module.  One aspect of the cron is the ability to do a probabilistic cron on request for people that don't have access to system crontabs.  The cron is run onEndRequest after the page is flushed.  This way page performance is not impacted.  The cron in onEndRequest does actually update the global state.  Saving the state for each task is redundant and this change would help calm things down.

belisoful:
I have been sensitized to operational order in working on __construct changes for behaviors.  To maximize functionality saveGlobals should be after onEndRequest.

ctrlaltca:
Thank you for the changes and sorry for the delay in reviewing them.
Looks like tests are failing, but that's due to a change in php-cs-fixer that needs to be fixed separately.

Issue #741, opened 2021-04-12
Behaviors should have a loader module/method from the application configuration


belisoful:
In working through how behaviors work, a TBehaviorsModule is likely needed as well as integration as a configuration thing.   I'm working on a TBehaviorsModule and then how to work it into the application configuration.

belisoful:
https://github.com/belisoful/prado/commit/a8558e33bc778529a5c26261a53e260ca42245a9#diff-7b3cab49d0e2e377e99523d2564d1854724af60299538119f262fee216231a5f

The behaviors module is complete.  Not sure if this needs to be folded not an application behavior.

belisoful:
Based upon the latest changes that Parameter Class Objects raise dyInit after setting of properties, here is a Behavior loader object that works as an application parameter.  This allows behaviors to be loaded as one of the first an app can do

https://github.com/belisoful/prado/blob/master/framework/Util/Behaviors/TBehaviorParameterLoader.php

ctrlaltca:
fixed in #764

PR #764, opened 2021-05-11
Adds TBehaviorsModule and TBehaviorParameterLoader with Unit Tests

belisoful:
There is an AdditionalBehaviors for TBehaviorsModule property that is an experimental feature and could use inspection.

This adds a new folder in Util for Behaviors that will be forthcoming.

This also corrects the TComponent unit test detaching events.


belisoful:
#741

Issue #765, opened 2021-05-12
Prado-cli needs to be cleaned up, Separate class files and namespaces, integration with PRADO


belisoful:
There are several Classes in prado-cli.  These should have their own defined PRADO classes and structures.  It is not possible to properly extend a cli class within Prado for Application Actions.  These prado-cli classes should be folded into Prado.  This refactor is for enabling an application action "cron".

Yii has a folder for Console.  They have an advanced console system.  At some point a better console system should be constructed.

belisoful:
Got things settled.  Shell command changes are needed to make Application specific commands, like for cron.

belisoful:
https://github.com/pradosoft/prado/pull/769

Completed with PR 769

PR #766, opened 2021-05-13
#765 Prado-CLI split into defined classes and files

### LINKED ISSUE #218: Improvement: Proper upgrade-mechanism for Prado
_From [google...@pcforum.hu](https://code.google.com/u/112363926660835410968/) on January 03, 2010 02:13:16_

Prado currently does't provide a proper and simple mechanism to upgrade an 
in-production site's code. If you try to just upgrade a codebase which is 
already active with Application::Mode set to "Performance", you'll run 
into trouble, as Prado will use partially or completely the old, pre-
parsed and cached versions of your templates, database schemas, etc. 

The document states that you can "fix" this by setting the 
Application::Mode first to "Debug" and then back to "Performance", which 
will clear the cache and let you proceed with your new templates, configs. 
However, this is actually not the case, as only those cache entries will 
get invalidated which are actually referred to in request while the 
application is running in Performance mode. That means, that unless you 
crawl your whole site while being in that mode, you will most likely still 
encounter old - now outdated - cache hits for templates in your code, 
which will cause all kinds of weird errors on your site. It's also 
ridicolous that you have to edit a text file (the application.xml) and set 
a switch back and forth to have your caches erased (which won't happen 
actually anyway, as just explained).

Therefore I propose a modification to the Prado TApplication to make such 
code/site upgrades easier. TApplication should be extended/modified to 
check for ex. the timestamp of the application.xml file at each startup, 
and if it detects any changes to the timestamp, it should clear all caches 
in use - therefore purging now probably outdated template, database, etc. 
entries.

The performance-penalty of this check should be practically zero, as Prado 
reads the application.xml file anyway at each startup. Also, there won't 
be unneccesary cache purges as application.xml file won't be ever changed 
unless there's some modification to the site itself - and if that latter 
doesn't involve modifying the application.xml file, then the file should 
be simple "touch"ed (last modification timestamp modified to current one) 
to force a proper purge of the Prado caches; this could be explained in 
the documentation. Purging all the caches will dispose of all the 
pagestates (if they're stored on server side) which have most likely 
become invalid anyway if templates have been changed - so it solves that 
problem too.

To accomplish the goals described above TApplication::onLoadState() should 
be extended with the code below:

``` php
  const CONFIG_TIMESTAMP_KEY = 'Prado:Application:ConfigTimeStamp'; // put 
this in class header

  $ts = intval($this->getGlobalState(self::CONFIG_TIMESTAMP_KEY, 0));
  $cts = intval(@filemtime($this->getConfigurationFile()));
  if ($ts!==$cts)
  {
    $app->setGlobalState(self::CONFIG_TIMESTAMP_KEY, $cts);
    $app->saveGlobals(); // save changes quick so we don't get redundant 
cache purge attempts while this is running
    foreach($app->getModules() as $module)
      if ($module instanceof TCache)
        $module->flush();
  }
```

This will check the timestamp of application.xml, and purge all 
application caches if the configuration file's timestamp has changed. 

Now if you upgrade your code on a production server, no more change 
between Debug and Performance modes (and the re-request of your pages) or 
manual deletion of caches neccessary. The only thing you have to do is 
to "touch" the application.xml file in the case your upgrade didn't affect 
that - and Prado will automatically reparse all your templates, database 
schemas, etc. and run properly with the updated code.

_Original issue: http://code.google.com/p/prado3/issues/detail?id=217_

belisoful:
This also resolves #218 by adding a flushcaches command.

This adds the "app" command to run commands internal to the app like cron.

Proper test cases can be written for the Shell operations now.


belisoful:
This is giving me some time to write the unit tests for what is coming.

PR #769, opened 2021-05-13
refactor CLI classes into individual files, create Shell directory

belisoful:
This also moves the TShellApplication from the framework directory to the Shell Directory


belisoful:
There was a bug in the ActiveRecordGen not having _overwrite variable and the functionality not working.  Apparently very few people used it and the bug went unnoticed.

belisoful:
TApplication has a method for getting modules by type.  This is useful in several other instances.

belisoful:
TApplication::GetModulesByType is ready

belisoful:
#765  The closing and reopening was to correct the TApplication::getModulesByType method

framework/Shell/Actions/TFlushCachesAction.php
```diff
@@ -0,0 +1,58 @@
...
+	public function performAction($args)
+	{
+		$app = null;
+		if (count($args) > 1) {
+			if (false === ($xml = $this->getAppConfigFile($args[1]))) {
+				return false;
+			}
+			$app = $this->initializePradoApplication($args[1]);
+		}
+		$app->onLoadState();
```
ctrlaltca, line 43:
```
PHP Fatal error:  Uncaught Error: Call to a member function onLoadState() on null in /var/www/htdocs/prado-master/framework/Shell/Actions/TFlushCachesAction.php:43

```
The directory parameter should be mandatory. 
Still, better check if $app is null before using it.

framework/Shell/Actions/TPphShellAction.php
```diff
@@ -0,0 +1,43 @@
...
+use Prado\Shell\TShellAction;
+
+/**
+ * Creates and run a Prado application in a PHP Shell.
+ *
+ * @author Wei Zhuo <weizhuo[at]gmail[dot]com>
+ * @package Prado\Shell\Actions
+ * @since 3.0.5
+ */
+class TPphShellAction extends TShellAction
```
ctrlaltca, line 23: TPphShellAction should probably be TPhpShellAction (typo)

framework/TApplication.php
```diff
@@ -698,6 +699,60 @@ public function getModules()
...
+	 * @since 4.2.0
+	 */
+	public function getModulesByType($type, $strict = false, $first = false)
+	{
+		$m = [];
+		foreach ($this->_modules as $id => $module) {
+			if ($module === null) {
+				[$moduleClass, $initProperties, $configElement] = $this->_lazyModules[$id];
+				
+				$match = ($moduleClass === $type);
```
ctrlaltca, line 719: Maybe this could be made easier and avoid the use of reflection?
```php
$match = $strict ? ($moduleClass === $type) : ($moduleclass instanceof $type);
```

framework/TApplication.php
```diff
@@ -698,6 +699,60 @@ public function getModules()
...
+	 * Returns a list of application modules of a specific class.
+	 * Modules that have not been loaded yet are loaded.
+	 * @param string $type class name of the modules to look for.
+	 * @param bool $strict should just the class be found or subclasses of the class
+	 * @param bool $first return the first module found, default false.
+	 * @return array|object list of loaded application module of a specific class
+	 * or the first object of a specific class
+	 * @since 4.2.0
+	 */
+	public function getModulesByType($type, $strict = false, $first = false)
```
ctrlaltca, line 712: Imho this method makes more things that what it should, making it a bit too complex.
I'd like it to return a list of modules ids, delegating to the caller the burden of calling `Prado::getApplication->getModule($id)` if it wants to use a module.
Since the caller would get an array of strings, we could also remove the $first parameter to make the method signature more strict on return types.
Please see the following comments.

framework/TApplication.php
```diff
@@ -698,6 +699,60 @@ public function getModules()
...
+						do {
+							if ($reflect->getName() === $type) {
+								$match = true;
+								break;
+							}
+						} while ($reflect = $reflect->getParentClass());
+					}
+				}
+				
+				if ($match) {
```
ctrlaltca, line 732: I'd avoid to make lazy modules load here.
We could return an array of module ids, and the caller would then call `Prado::getApplication->getModule($id)` on the module if it needs to access it

framework/TApplication.php
```diff
@@ -698,6 +699,60 @@ public function getModules()
...
+				}
+			} elseif (get_class($module) === $type || (!$strict && ($module->isa($type)))) {
+				if ($first) {
+					return $module;
+				} else {
+					$m[] = $module;
+				}
+			}
+		}
+		if ($first) {
```
ctrlaltca, line 749: I's remove the $first parameter, making this method always return an array (possibly empty) of module ids.

framework/Shell/TShellInterpreter.php
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * TShellInterpreter Class
+ *
+ * Command line interface, configures the action classes and dispatches the command actions.
+ *
+ * @author Wei Zhuo <weizhuo[at]gmail[dot]com>
+ * @package Prado\Shell
+ * @since 3.0.5
+ */
+class TShellInterpreter
```
ctrlaltca, line 24: What do you think of making TShellInterpreter a subclass of TShellApplication?
It would reuse TShellApplication::_actionClasses related methods, creating the actual classes inside  TShellInterpreter::run()

framework/Shell/TShellInterpreter.php
```diff
@@ -0,0 +1,106 @@
...
+
+	/**
+	 * Print command line help, default action.
+	 */
+	public function printHelp()
+	{
+		TShellInterpreter::getInstance()->printGreeting();
+
+		echo "usage: php prado-cli.php action <parameter> [optional]\n";
+		echo "example: php prado-cli.php -c mysite\n\n";
```
ctrlaltca, line 100: This example is broken and can be removed.

framework/Shell/Actions/TAppHelpAction.php
```diff
@@ -0,0 +1,36 @@
...
+	protected $optional = [];
+	protected $description = 'outputs the CLI Application help';
+	
+	/**
+	 * @param array $args parameters
+	 * @return bool
+	 */
+	public function performAction($args)
+	{
+	}
```
ctrlaltca, line 35: I don't fully get what this is meant to do, is it just a stub to be used by cli applications?

framework/Shell/Actions/TAppAction.php
```diff
@@ -0,0 +1,57 @@
...
+		}
+		if (false === ($xml = $this->getAppConfigFile($args[1]))) {
+			echo "** Application '$args[1]' was not found. \n";
+			return false;
+		}
+		$app = $this->initializePradoApplication($args[1]);
+		TShellInterpreter::getInstance()->clearActionClass();
+		TShellInterpreter::getInstance()->addActionClass('Prado\\Shell\\Actions\\TAppHelpAction');
+		foreach ($app->getShellActionClasses() as $actions) {
+			TShellInterpreter::getInstance()->addActionClass($actions);
```
ctrlaltca, line 48: This class basically resets the shell intepreter, loads a fixed set of actions and re-runs the shell interpreted.
I guess this would be somewhat used by an app to add its own actions to the shell interpreter, but how to do it?

ctrlaltca (changes requested):
Very nice refactoring, i have a suggestion on TApplication::getModulesByType() and i'm a bit lost on the usage of the "app" action, please have a look at the comments

belisoful:
Briefly.  Many of the issues have been remaining for a while.  Now is a great time to address them.

A ShellAppAction is different from an ShellAction in that it is an application specific command rather than a Prado command.

When calling 

`prado-cli app /my_app `

the ShellAppAction allows for internal application commands, eg.

`prado-cli app /my_app cron`

and

`prado-cli app /my_app cron tasks`

These commands are implemented by an application's module(s) rather than from a Prado command

framework/Shell/Actions/TAppAction.php
belisoful: That was exactly the issue I was running into.  I decided to not worry about it *yet* and just do the refactor to see what ideas precipitated from it.  the Prado-cli would need to be redesigned.

eg. the first parameter would need to be the directory, or better yet have a -d or --directory input flag, otherwise the cli checks the current directory or 'protected' for the application.

I didn't know what you thought about re-doing the Shell and what metrics should be used.

framework/Shell/Actions/TFlushCachesAction.php
belisoful: Agreed.  It was really weak sauce to begin with, I didn't both proofing it, obviously.  I didn't want to spend eternity on this refactor even though it's looking like I might have to spend some time making it reasonable.  lol.

framework/Shell/Actions/TPphShellAction.php
belisoful: indeed. I changed it on one version but it wasn't the version that was checked in.  Changing now.

framework/Shell/TShellInterpreter.php
belisoful: I thought that might be the case but was waiting for your comment on this.

framework/Shell/Actions/TAppHelpAction.php
belisoful: This is a stub, the actual display of help happens in the TShellInterpreter::printHelp through TAppAction when no valid option is found.  This apparently doesn't need to display the help, as it is automatically displayed when help is triggered for the application.

framework/Shell/TShellInterpreter.php
belisoful: brilliant.

belisoful: I love the idea but there needs to be a directory specified by each command is not standardized.  This could get rid of some of the hacks for making application commands. the application could just MAKE a command without it being an app command.  The only way to get application specific commands is by having the directory to begin with.

framework/Shell/Actions/TAppHelpAction.php
belisoful: I added return false, and documented that behavior

framework/Shell/Actions/TFlushCachesAction.php
belisoful: updated

framework/TApplication.php
belisoful: I'm matching the output of GetModules.

framework/Shell/TShellInterpreter.php
ctrlaltca: Got it

ctrlaltca (approved):

ctrlaltca:
Wonderful, thank you!

Issue #737, opened 2021-04-03
Should Theme Skin Templates not have their classes and attributes validated?


belisoful:
Currently, Theme Skins are treated as standard templates and as such are validated for classes and attributes.  Such strictness is useful for templates that are rendered.  Is such strictness necessary for Skinning?

Maybe we should turn off class/attribute validation for theme Skin templates.  If a custom Control class is deleted (or theme reused from another site), Should the skin throw an error because it couldn't find the class?

I propose that Theme Skins have their class/attribute validation turned off so classes in the skin that don't exist don't throw an error.

I already have this change coded up, but I'll wait for discussion until posting.

belisoful:
Application of a skin is where errors would take place rather than on skin validation.  This seems like preferred behavior.

The Template acts as a storage of skin parameters, rather than it being rendered.  As such, validation is less important.

ctrlaltca:
If I get it correctly you mean that when you define a skin that applies to a few classes/controls, you currently receive a parsing errors on the skin template if one of the classes is not available in the application.
To make the same skin it reusable in multiple apps, you currently need to remove from the skin all the definitions for classes that are not available in that specific app.
Ignoring them would make the skin more reusable. If a skin contains errors, they will still be reported when the skin gets appllied on a control.
This looks like a good idea.

belisoful:
I think you got it. yes.

belisoful:
merged.

PR #773, opened 2021-05-15
incremental update Corrections, speedups-caching, and Themes using TSkinTemplate without validation

belisoful:
This is a general update to

- TMapRouteBehavior- changes the public variable to a private variable with getter/setter. It adds the method to the owner but that shouldn't be an issue.
- Adds a Javascript extension to Number and String adding the method px() that adds "px' to the end of a number or string.  The string checks if px is at the end and if not, adds it.  This should make conversion from numbers and strings to css pixel sizes more standardized and uniform.
- TApplicationComponent adds caching the fx events in the application cache when in normal or performance mode.
- TTemplate adds attribute validation property.
- TSkinTemplate sets TTemplate attribute validation to false
- TThemes use TSkinTemplate rather than TTemplate for skins #737

### DIFF
--- a/framework/TApplicationComponent.php
+++ b/framework/TApplicationComponent.php
@@ -10,6 +10,8 @@
 
 namespace Prado;
 
+use Prado\TApplicationMode;
+
 /**
  * TApplicationComponent class
  *
@@ -33,6 +35,7 @@ namespace Prado;
  */
 class TApplicationComponent extends \Prado\TComponent
 {
+	public const APP_COMPONENT_FX_CACHE = 'prado:applicationcomponent:fxcache';
 	/**
 	 * TApplicationComponents auto listen to global events.
 	 *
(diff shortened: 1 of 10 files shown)


belisoful:
FYI. I figured out what happened to merge without intending to... the GihHub website "merge resolution tool" chose the incorrect merge direction that was not specified by the PR.  Both times I used the tool and it went the unintended direction.  I was going to equal the branch changes and the website merge resolution tool had, in a very tiny and out of the way location, the direction of the merge... and it wasn't changeable.  That's what was strange.   that's what happened.  staying away from that method now. shouldn't happen again.

framework/TApplicationComponent.php
```diff
@@ -43,6 +46,40 @@ public function getAutoGlobalListen()
 		return true;
 	}
 	
+	/**
+	 * This caches the 'fx' events for PRADO classes in the global state
+	 * @param object $class
+	 * @return string[] fx events from a specific class
+	 */
+	protected function getClassFxEvents($class)
+	{
```
ctrlaltca, line 55: Wow, I see a lot of optimizations going on here.
The method is quite complex, but i didn't find any case where it fails.
I suppose it took a lot to get it so well done.

ctrlaltca (approved):

framework/TApplicationComponent.php
belisoful: It has been tested and tested and tested, yes.  The final touch was to run on Normal and Performance Mode only.  I figured that there needed to be a way to speed this up more.

PR #774, opened 2021-05-17
Major Behavior Update

belisoful:
- Corrected the TApplicationComponent php doc (changed 'global state' to 'application cache')
- Added TPage*Behaviors for top anchor, no cache, and globalization charset
- Added the TParameterizeBehavior (This is a MAJOR component with significant implications)
- added the behaviors to classes.php
- added TParameterizeBehavior exceptions to messages.txt
- Added TTimeZoneParameterBehavior
- Unit Tests for all 5 new behaviors.

These classes affect the base application project functionality.  These functionalities are built into masterlayout and demo page of the base application project.  These behaviors can be installed on the base application project with TBehaviorsModule or TBehaviorParameterLoader. fyi.

**TParameterizeBehavior** is a _significant_ class moving forward.  Any object property can be initialized from any application parameter.  eg.  This class can take an app param for Title and map it to TPage.Title property.  It can take an app param for Theme and map it to the page theme.

The **TTimeZoneParameterBehavior** adds TimeZone methods to TApplication via behavior.  It also sets the time zone from a specified application parameter.


belisoful:
TParameterizeBehavior can be used to set WebControl defaults from application parameter.

eg. TGravatar Rating and style can be defined by app parameter (as well as through Skinning). TparameterizeBehavior can be used to make new object defaults regardless of skin support.

ctrlaltca (approved):

PR #781, opened 2021-05-28
TPriorityList bug fix with unit test, TApplication::getModulesByType correction, TEventContent & TPageLoadTime (addition with unit test)

belisoful:
TPriorityList bug - cache did not update when exists but empty, fixed with unit test, caching performance speed-up

TApplication::getModulesByType - when setting modules to null, null modules may not be a sign of lazy load modules

TEventContent is an experimental control for allowing composer PRADO package modules to fill in controls, like navigation.  A global event is raised and the handlers fill in controls (like list items).

TPageLoadTime measures the time from starting the request to the tag render time.


framework/Util/TDbParameterModule.php
```diff
@@ -233,14 +233,19 @@ protected function createDbTable()
 		$db = $this->getDbConnection();
 		$driver = $db->getDriverName();
 		$autoidAttributes = '';
+		$autotype = 'INTEGER';
 		$postIndices = '; CREATE UNIQUE INDEX tkey ON ' . $this->_tableName . '(' . $this->_keyField . ');' .
 		($this->_autoLoadField ? ' CREATE INDEX tauto ON ' . $this->_tableName . '(' . $this->_autoLoadField . ');' : '');
 		if ($driver === 'mysql') {
```
ctrlaltca, line 239: use switch() instead?
The default branch could either default to mysql or throw an exception.

framework/Web/UI/WebControls/TPageLoadTime.php
```diff
@@ -0,0 +1,54 @@
...
+	 */
+	public function renderContents($writer)
+	{
+		$writer->write(round(microtime(true) - $_SERVER["REQUEST_TIME_FLOAT"], 5) . Prado::localize($this->getSecondAppendix()));
+	}
+	
+	/**
+	 * @return string the string that is appended to the time.  default 's' for seconds.
+	 */
+	public function getSecondAppendix()
```
ctrlaltca, line 42: Suffix?

ctrlaltca (approved):
Looks good! two minor comments, not blocking.

framework/Util/TDbParameterModule.php
belisoful: good point

framework/Web/UI/WebControls/TPageLoadTime.php
belisoful: changed

framework/Util/TDbParameterModule.php
belisoful: corrected

PR #785, opened 2021-06-06
Shell Colorization, TCallChain bug fix

belisoful:
All shell output goes through the TShellWriter and TOutputWriter rather than echo.  This is for future developments.


ctrlaltca (approved):
Another great addition!
In the future we may want to standardize the output of errors (lines starting with "** ").
Also, actions are calling $this->_outWriter->... without null-checking it before, but i guess that's fine since the actions should only be created by TShellInterpreter that ensures it is set.

belisoful:
I agree that some kind of template for errors should be formed.  I am working towards that direction.  the output writer will always be present due to being a shell command, correct.

PR #786, opened 2021-06-11
Composer Integration: module bootstrap in app configuration, TPluginModule, IDbModule interface

belisoful:
Composer packages can use composer.json `"type": "prado4-extension"`.   The composer.json also can have an ['extra']['bootstrap'] Class.  eg:

`"extra": { "bootstrap": "PradoComposerBase\\MainModule" }`

The way Prado composer extensions are instanced are almost like normal modules but without the "class" attribute.  The id of the module is the name of the composer package.

Here is the first example and template Prado composer extension.  This should probably be transferred to be under the Prado vendor rather than me.
https://github.com/belisoful/prado-composer-base

Setting the Application Parameter 'PluginContentId' will set the composer extension Pages TContent ID.

The TPluginModule will automatically check for a 'Pages' folder and add it to the TPageService::onAdditionalPages handler.  It also checks for errorMessages.txt and adds it to the TException message files.

TControl can find a a composer extension from its own directory path.

Also the prado-cli autoload sequence order is reversed.  the application is loaded first, then the repo.


framework/TApplicationConfiguration.php
```diff
@@ -234,6 +250,60 @@ protected function loadPathsXml($pathsNode, $configPath)
...
+			$plugins = $cache->get(self::COMPOSER_INSTALLED_CACHE);
+			if ($plugins !== null) {
+				return $plugins;
+			}
+		}
+		$dependencies = new TChainedCacheDependency();
+		$listDeps = $dependencies->getDependencies();
+		$plugins = [];
+		foreach (ClassLoader::getRegisteredLoaders() as $vendorDir => $loader) {
+			echo($vendorDir . "\n");
```
ctrlaltca, line 273: Remove debug, or change it to a Prado::trace() if useful

framework/TApplicationConfiguration.php
```diff
@@ -234,6 +250,60 @@ protected function loadPathsXml($pathsNode, $configPath)
...
+			$file = $vendorDir . DIRECTORY_SEPARATOR . 'composer' . DIRECTORY_SEPARATOR . 'installed.json';
+			$manifests = json_decode(file_get_contents($file), true, 512, JSON_THROW_ON_ERROR);
```
ctrlaltca, line 275: JSON_THROW_ON_ERROR requires php 7.3
Since 7.2 is already EOL, we can just require php >= 7.3 for prado 4.2

framework/Util/TDbPluginModule.php
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * @author Brad Anderson <belisoful@icloud.com>
+ * @link https://github.com/pradosoft/prado
+ * @license https://github.com/pradosoft/prado/blob/master/LICENSE
+ * @package Prado
+ */
+	
+namespace Prado\Util;
+
+use Prado\Util\XTPluginModule;
```
ctrlaltca, line 14: This class doesn't exists, is it a leftover?

framework/Util/TPluginModule.php
```diff
@@ -0,0 +1,182 @@
...
+
+/**
+ * TPluginModule class.
+ *
+ * TPluginModule is for extending PRADO through Composer packages.  This installs
+ * its own Pages, where available, and its own error message file.
+ *
+ * Plugin pages should implement their *.page with the following page code:
+ * <code>
+ *	<com:TContent ID=<%= Prado::getApplication()->getParameters()->itemAt('PluginContentId') %>>
```
ctrlaltca, line 27: You can probably use a property tag here:
```html
<com:TContent ID=<%$ PluginContentId %>>
```

framework/Web/UI/TControl.php
```diff
@@ -346,6 +351,32 @@ public function getSourceTemplateControl()
...
+	public function getPluginModule()
+	{
+		if ($this->_pluginmodule === false) {
+			$this->_pluginmodule = null;
+			foreach ($this->getApplication()->getModulesByType('Prado\\Util\\IPluginModule') as $id => $module) {
+				if (!$module) {
+					continue;
+				}
+				$reflect = new ReflectionClass(get_class($this));
+				$folder = $reflect->getFileName();
```
ctrlaltca, line 369: Move these two lines out of the foreach loop to avoid executing them multiple times

ctrlaltca (changes requested):
This is awesome! I can-t stand testing it.
I have a few minor feedbacks on the code.

ctrlaltca:
> Here is the first example and template Prado composer extension. This should probably be transferred to be under the Prado vendor rather than me.
> https://github.com/belisoful/prado-composer-base

I've added yourself to the maintainers team, you should be able to create and migrate the repo

framework/TApplicationConfiguration.php
belisoful: oops.  my bad.  I'll get on this.

framework/Util/TDbPluginModule.php
belisoful: This is a leftover.  I'll remove it.

framework/Util/TPluginModule.php
belisoful: For some strange reason, the property tag did not work.  I'll have another look at this.

framework/TApplicationConfiguration.php
belisoful: I'm so glad you like this composer integration PR.  It's sorely needed for the community.  This is modeled on Yii.

framework/TApplicationConfiguration.php
belisoful: Perfect.

belisoful:
I really hope this PR will energize the Prado community to be more active and produce their own composer modules for Prado.

framework/Util/TPluginModule.php
belisoful: oh right.  when using a property tag rather than = expression tag this is the error:
Invalid template syntax: TContent.ID can only accept a static text string

I don't know what the intended behavior is of the template and parsing, but it works as is without the property tag.

belisoful: given your experience, can you clarify?  I'm looking into the template parsing to see what is going on.

belisoful: https://github.com/belisoful/prado/blob/3f3ed1b1ba854d4634f6cf311bc427a481ade33e/framework/Web/UI/TTemplate.php#L846

This is why.  Adding an ID/SkinId to be keyed from the parameter needs to be coded into the template.  I have a fix. coming.

framework/TApplicationConfiguration.php
belisoful: This probably doesn't need to be JSON_THROW_ON_ERROR because the installed.json *should* always work

belisoful: Is this a worthy dependency? in your view?

ctrlaltca: Keep it, people running EOL php versions already have worst problems

Issue #761, opened 2021-05-05
Cron Module for offline app tasks


belisoful:
One of the major components that seems to be missing from PRADO is a cron module.  This would ideally be run from the system crontab via Prado-cli every minute.  The app then decides whether to run any tasks.  Each Prado crontab run takes the same overhead as a request, which should be acceptable to run once a minute to check for any pending cron tasks.

Tasks should be scheduled via cron style * * * * * * string.  This should be as advanced as possible with scheduling options. .eg. W, L and # should be supported, in the proper places (of course).

I've been working on this for a week now.  The scheduler is a lot of work.  Forking produced inconsistent results so there won't be parallel tasks.  There is an option to run cron on request by probability for those who don't have access to system crontabs.  Each task is a defined class.  There is an extension for DbCron for logging tasks and taking on real time tasks.

Real time cron tasks might be sending an email offline, outside of the request application stack.  The email can be scheduled for next run.

belisoful:
The unit test for the cron schedule is massive.  MANY test cases.  So far its adds 1300 tests (almost 40% of the exists test cases) and its not even close to done.

belisoful:
The tuning on this is a lot of work.

PR #787, opened 2021-06-14
#761 TCronModule, unit tests, TTimeScheduler

belisoful:


framework/Util/Cron/TCronTask.php
```diff
@@ -0,0 +1,247 @@
...
+	public function getNextTriggerTime()
+	{
+		$s = $this->getScheduler();
+		return $s->getNextTriggerTime($this->_lastexectime);
+	}
+	
+	/**
+	 *
+	 */
+	/*public function getPreviousTriggerTime()
```
ctrlaltca, line 196: Is this a leftover to be removed or you want to decomment and keep it?

framework/Util/Cron/TTimeScheduler.php
```diff
@@ -0,0 +1,856 @@
...
+	}
+	
+	
+	/**
+	 * Given which hours are valid for the parsed time stamp, this generates
+	 * an array of valid hours.  Each hour is available as the key to the array
+	 * and the value contains true or false whether the hour meets the cron
+	 * time stamp
+	 */
+	protected function geHoursArray()
```
ctrlaltca, line 369: This should probably be named getHoursArray(), with and added "t"

framework/Util/Cron/TTimeScheduler.php
```diff
@@ -0,0 +1,856 @@
...
+	
+	/**
+	 * Returns the number of days in a given month and year, taking into account leap years.
+	 *
+	 * @param $month: numeric month (integers 1-12)
+	 * @param $year: numeric year (any integer)
+	 */
+	public function days_in_month($month, $year)
+	{
+		return $month == 2 ? ($year % 4 ? 28 : ($year % 100 ? 29 : ($year % 400 ? 28 : 29))) : (($month - 1) % 7 % 2 ? 30 : 31);
```
ctrlaltca, line 391: This could be done with
```php
cal_days_in_month(CAL_GREGORIAN, $month, $year)
```
But i guess you avoided that method since it requires the calendar module.
As an alternative you can use this:
```php
date('t', mktime(0, 0, 0, $month, 1, $year)); 
```
I'm not sure if this really makes any difference in terms of performance, so feel free to use the approach that you prefer.

framework/Util/Cron/TTimeScheduler.php
```diff
@@ -0,0 +1,856 @@
...
+	}
+	
+	
+	/**
+	 * This calculates the next trigger time for the schedule based on the parameter
+	 * time.  If no parameter time is given, the current system time is used.
+	 * @param mixed $priortime
+	 * @return numeric the unix time of the next trigger event.
+	 */
+	/*public function getPreviousTriggerTime($priortime = false)
```
ctrlaltca, line 544: Is this needed?

ctrlaltca (approved):
Looks good, with minor comments, but feel free to keep it as it is if you want.
At first I didn't like the way tasks are defined as simple strings that gets eval()uated (I would prefer a standard callable array), but seeing how you made them configurable in the application.xml I changed my mind.

framework/Util/Cron/TTimeScheduler.php
belisoful: ah yeas.  true

framework/Util/Cron/TCronTask.php
belisoful: I wanted to implement it, but that is a LOT of work just to get the NextTriggerTime working.  phew!   took me a week and a half just to get the Next working.  The previous is optional..  so, I'll remove it for now.

belisoful:
There are two kinds of tasks,  one is a class, the other is as an evaluated string from configuration.  The DB portion is next, so keep that in mind.  The TDbCronModule can be found in my Prado master if you want to check out the whole thing.

framework/Util/Cron/TCronTask.php
belisoful: If you want to implement it, please, by all means, go for it.  It should be simply duplicated code from the NextTriggerTime except backwards.   The test units are the real monster.

framework/Util/Cron/TTimeScheduler.php
belisoful: good eye on this.

framework/Util/Cron/TCronTask.php
ctrlaltca: I don't see a point for it in a simple cron scheduler. Feel free to drop it

framework/Util/Cron/TTimeScheduler.php
belisoful: No.  though if someone in the community wants to write it, for completion, then it would be a gift.  The unit test is huge.

framework/Util/Cron/TTimeScheduler.php
belisoful: 100.000 times and here are the results:
2.5287721157074  for cal_days_in_month
1.4125881195068  for date takes
0.017313957214355  for current solution

ctrlaltca:
Feel free to squash and merge by yourself when you are done with this one

belisoful:
I do think we are there.  This PR has been in the works, technically, for over 7 years.  I'm going to have another round of look through, mainly the comments and errors.  This uses a lot of seemingly typically unused features of PRADO, behaviors, global events, shell, dynamic events.

I'm disliking the way cli finds the app directory and could be updated with composer (in mind) to a more fixed relationship.  Where in the usual, there would be no app directory specified as a parameter.  It can be found automatically, by default.  This would allow Application actions to be integrated with hard coded actions and simplify the arguments.  I'd like the cron shell to be cleaner than it is this moment.

At some point, how about a 4.2.0-dev tag?  the prado-composer-extension code can use that as the minimum version. and people can start working that code to make extensions?  I know I'd like to start releasing some of my other code as extensions.  Making this PRADO ecosystem of extensions has been a vision for a while now.

belisoful:
Let's merge the TDbCronModule before tagging the dev. which is the next pr

PR #788, opened 2021-06-16
TDbCronModule with unit tests

belisoful:
After this PR let's do the 4.2.0-dev tag.


ctrlaltca (approved):

ctrlaltca:
Merged and tagged 4.2.0-alpha1
I could use 4.2.0-dev ad a tag because packagist skips versions containing "dev"

PR #795, opened 2021-07-13
TAuthorizationRule Update & TPageConfiguration extendability

belisoful:
- Added IPriorityItem to classes.php

- TAuthorizationRuleCollection is upgraded from TList to TPriorityList so the rules can implement IPriorityItem
- TAuthorizationRule can be constructed without parameters.
- TAuthorizationRule has setters for all properties for instancing with properties
- TAuthorizationRule is serializable for storage of the rule, zappable parameters
- TauthorizationRule implements IPriorityItem so rules can be prioritized when coming from different sources (permission file, page configuration, and database).  eg. Deny All can be prioritized as LAST even when there are other rules yet to be added.  Rules are no longer ordered by configuration order when the rule has a priority.
- TauthorizationRule Unit Test update for setters and serialize/unserialize object self similarity check

- TPageConfiguration enables TAuthorizationRules to have priorities
- TPageConfiguration adds dynamic events for extending the use of Page Configurations to other modules (loadXML, loadPHP, and apply dynamic events). [this is mainly for TPermissionsManager to have different permissions and roles specified per directory, useful for any module that wants to extend TPageConfiguration]

- IsUserAllowed adds an Extra method parameter for passing data from the calling method to the authorization rule.  There is a new AuthorizationRule in Permissions that allows when the user own the data.  eg.  blog_update, if the user owns the blog content, the update is allowed.  this requires passing the owner of the data to the isUserAllowed method.  Yii has other ways of dealing with this that could be (re-)implemented later.


ctrlaltca:
These changes looks fine from a quick review. Did you close this PR by accident?

belisoful:
I want to change the TAuthorizationRule so it doesn't have the setter methods.  That change is unnecessary.  I closed this out of respect that the changes could be made more minimal.  I have an update over the next day or so.

belisoful:
I really like that TPageConfiguration adds dynamic events for extending.

Issue #783, opened 2021-05-29
Revamp of CLI, colorize, add cursor controls, localize text


belisoful:
The revamped cli should initiate the application from the current directory (or protected), or from the vendor bin, or otherwise be directed to the application directory with the option "-d" or "--directory=".   This standardizes how the application directory is set with the commands.  

Once the application starts up, application specific shell commands are loaded as well as non-application specific.

This removes the ad-hoc "app" command for running application specific commands.

This would allow PRADO to localize console text as well.

belisoful:
For localizing text, this is helpful to acknowledge:  https://www.baeldung.com/linux/locale-environment-variables

belisoful:
The TShellWriter can be updated from here with error formatting relatively easily.  Other platforms have such larger structures for formatting shell output.  I wanted to have some discussion about this before moving on it.

belisoful:
I borrowed a few ideas from Yii for implementing the new CLI.  Actions has sub-actions for implementing specific functionality.  It implements the same way Yii does it with "route/action" mapping to method "action" followed by the route action.  so "cron/index" is mapped to "actionIndex($args)"

belisoful:
I have a behavior for forcing users to log in to use the shell There are other places that CLI improperly uses the session that need attention.  I'm looking at this and other aspects.  This upgrades cron shell as well.  There is now a shell action for getting and setting parameters to the DB.

belisoful:
This is on its way.  I'm just wrapping up the unit test for TPermissionsManager.  The CLI cannot stay in its current condition given the new update I have being processed.

Issue #26, opened 2013-09-14
[790] TFileUpload doesn't work with active controls - maybe add TActiveFileUpload?


ctrlaltca:
_From [haertl.mike@gmail.com](https://code.google.com/u/haertl.mike@gmail.com/) on October 17, 2008 23:41:04_

Using a TFileUpload control and submitting the form with a TButton works.
However, doing the same with a TActiveButton does not as $_FILES is empty
in the handler. Adding some kind of active-aware version of TFileUpload
would help.

_Original issue: http://code.google.com/p/prado3/issues/detail?id=25_

PR #796, opened 2021-07-15
#783 Shell modernization for Composer, Cron Shell refactor, Db Parameter Shell, 

belisoful:
Unified the __destruct parent-child chains to ensure unlisten is called.

Shell adds action/routes, application options, application optionAliases, command specific options and aliases, implemented the application stack on the shell application.  Removed AppAction hack.

Cron is also updates a field from UserId to UserName for consistency on the platform

This also adds IPermissions, TPermissionEvent, and TUserOwnerRule under the Permissions subsystem.  These are used by other parts of PRADO.  I don't want to waste time removing these parts just to do an update only to add them back in.

Look at the individual commits starting with Merge #26 on Jul 14th.


belisoful:
My humble apologies for the size of this PR.  It was meant to be split in two but that was a LOT of redundant work.

ctrlaltca (commented):
Hi, trying this PR I'm not able to use prado-cli from an application anymore:
```
#./protected/vendor/bin/prado-cli 
Application runtime path '/var/runtime' does not exist or is not writable by Web server process.
#0 ${PradoFramework}/TApplication.php(307): Prado\TApplication->resolvePaths()
#1 /var/www/htdocs/prado-master/bin/prado-cli.php(41): Prado\TApplication->__construct()
#2 /var/www/htdocs/prado-master/bin/prado-cli(4): include('/var/www/htdocs...')
#3 {main}
```

bin/prado-cli.php
```diff
@@ -21,22 +22,21 @@
...
-
 restore_exception_handler();
 
-//register action classes
-TShellInterpreter::getInstance()->addActionClass('Prado\\Shell\\Actions\\TAppAction');
-TShellInterpreter::getInstance()->addActionClass('Prado\\Shell\\Actions\\TFlushCachesAction');
-TShellInterpreter::getInstance()->addActionClass('Prado\\Shell\\Actions\\TPhpShellAction');
-TShellInterpreter::getInstance()->addActionClass('Prado\\Shell\\Actions\\TActiveRecordGenAction');
-TShellInterpreter::getInstance()->addActionClass('Prado\\Shell\\Actions\\TActiveRecordGenAllAction');
+$app_dir = dirname(__DIR__, 4);
```
ctrlaltca, line 27: This seems to be the line causing prado-cli trying to load the application from /var in my error

framework/Shell/TShellWriter.php
```diff
@@ -184,13 +184,100 @@ public function writeError($text)
...
 	
+	/**
+	 * @param string $str the string to ANSI format.
+	 * @param string|string[] $attr the attributes to format.
+	 * @param mixed $len
+	 * @param mixed $pad
+	 * @param mixed $place
+	 * @return string $str in the format of $attr.
+	 */
+	public function pad($str, $len, $pad = ' ', $place = STR_PAD_RIGHT)
```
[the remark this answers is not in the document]
belisoful: The method is colorized enabled.  PHP function treats colorization as not control text but printable text.

bin/prado-cli.php
belisoful: Do you have Prado in an application via composer?  It sounds like a machine specific installation outside composer.  The base directory being presumed is:
/directory_to_app/web_app/protected/vendor/prado-cli which links to
/directory_to_app/web_app/protected/vendor/prado/pradosoft/bin/

Hence shaving off 4 directories to get the app directory.  Outside the bin, -d option is for setting the app_dir.

Should this be checking another directory for being the app dir?  It sounds like it should, given your setup.

belisoful: After reviewing Yii and other platforms, They require an application configuration to run.

This PR changes the assumption that Prado doesn't need an application to run.  Before this PR, Prado was able to perform some base level functionality effectively outside the application framework.  The main functionalities of the shell required an application directory path.

This PR internalizes the application directory path based upon composer and automatically starts up the app config to provide the functionality.  All functionalities remove the requirement of providing an app directory path because it is built in.

This case covers most installations.  I saw your comment above with the error and it does seem that your Prado-master is operating outside the context of composer (and for development purposes, rightly so).

In such situation a "-d=/path/to/app_config" option is needed to load the application.  Removing the app config directory into an option standardizes the commands and creates the application context prior to running the interpreter/application.  This was also to remove the hack "app action" for application actions that needed a loaded application context.

I hope this assumption is not too far out there.   Yii has a similar implementation of their console.   This PR is modeled on the Yii command line so we can capture some people from that bankwagon and make it easy for ppl to grasp coming from that background.

belisoful: Should the error (no application in provided directory, try -d to specify your own directory) be more graceful?  That is what I am thinking.

belisoful: I think checking the current directory (and possible ./protected) for valid app_directory would also be a good thing to have as well

belisoful: I'm putting an app config check function in prado-cli.  It checks the composer vendor directory, and if the application configuration is not found (like in your instance), it checks the current working directory C.W.D., and if not found there, checks the CWD "/protected" directory.  This should cover your instance.

If not found in these three default directories, throws a proper Prado Shell Error telling the use to specify the app config directory with option "-d="

If the user specified the directory with "-d" but no application is found, checks the 'protected' sub-directory.  if not found throws a proper Prado Shell Error.

This is being tested and will check in soon.  In the mean time, you can use "-d=." to specify the current directory as the application to operate from.  The checkin will eliminate the need to specify the app directory in your instance (where the app directory is the CWD)

ctrlaltca: >  It sounds like a machine specific installation outside composer.
You are right, sorry about this. I forgot i had some messy symlink structure to test prado master and PRs from the app.
Theoretically someone could cause the same problem installing prado globally using composer (composer global require), but I don't think that's a common (if even correct) usage at all.

> Should the error (no application in provided directory, try -d to specify your own directory) be more graceful? That is what I am thinking.

Yep, we should definitely provide an error message if the app directory is not found. The php backtrace for `Prado\TApplication->resolvePaths()` is just too cryptic.
Previously there used to be an "unable to load Prado application from directory xxx" error message

> I'm putting an app config check function in prado-cli. It checks the composer vendor directory, and if the application configuration is not found (like in your instance), it checks the current working directory C.W.D., and if not found there, checks the CWD "/protected" directory. This should cover your instance.

That would be awesome, thank you

belisoful: Fantastic.  I wasn't sure how to handle a Prado install from a repo but it's been figured out here.  I'll be posting the update soon.

belisoful: don't forget to try "-d=/tmp" to trigger an error so you can see what that looks like.

ctrlaltca:
Without any action parameter i get an error:
```
# ./protected/vendor/bin/prado-cli                                
  Command line tools for Prado 4.2.0.
[Notice] Undefined offset: 0 (@line 123 in file /var/www/htdocs/prado-master/framework/Shell/TShellAction.php).
#0 ${PradoFramework}/Shell/TShellAction.php(123): Prado\Prado::phpErrorHandler()
#1 ${PradoFramework}/Shell/TShellApplication.php(173): Prado\Shell\TShellAction->isValidAction()
#2 ${PradoFramework}/TApplication.php(388): Prado\Shell\TShellApplication->runService()
#3 ${PradoFramework}/Shell/TShellApplication.php(102): Prado\TApplication->run()
#4 /var/www/htdocs/prado-master/bin/prado-cli.php(86): Prado\Shell\TShellApplication->run()
#5 /var/www/htdocs/prado-master/bin/prado-cli(4): include('/var/www/htdocs...')
#6 {main}
```

framework/Shell/Actions/TPhpShellAction.php
```diff
@@ -16,34 +16,34 @@
...
-			$this->getWriter()->flush();
-			if ($loaded === false) {
-				return true;
-			}
-		}
-
+		$this->getWriter()->flush();
+		
 		$shell = new \Psy\Shell();
 		$shell->setBoundObject($this);
```
ctrlaltca, line 44: This used to bound $this to Prado::getApplication(), wether it now binds to the TPhpShellAction object.
Even if it's not related/caused by your PR, can you please restore it?

belisoful: I'll get on this.  Good eyes.

belisoful: The fix for no action parameter error (stated above) is a one line correction also.

belisoful: I didn't know how PsyShell worked or I would have bumped the bug earlier.

as an aside, the TPermissionManager is looking nice and firm.  I'm working on a shell action for permissions.

PR #806, opened 2022-03-24
Adapt for compatibility with PHP 8.1

zendre4:
Hello, I found this bug when I try to migrate in master version with PHP 8.1 .

uniqid don't allow null at parameter anymore...


ctrlaltca:
Nice catch, thank you!

zendre4:
thanks to you for the job !

PR #811, opened 2022-05-31
Update TSqlCriteria.php for fix phpdoc

zendre4:
The phpdoc comment was not correct. the functions can accept array an ArrayAccess but only Array access as defined on the comment. So my IDE inform me than it's not correct when I pass an array in the functions


ctrlaltca:
Can you please fix the order of types? Thank you

zendre4:
Done

ctrlaltca:
Merged, thank you!

Issue #815, opened 2022-11-14
T*CronModule issues


szroman:
Hi,

At the beginning I would like to thank you for creating CronModules, nice job.

Issues:
1. New tasks are always executed during the first run of prado-cli cron/run, regardles of scheduled time of execution.
2. Expired tasks (Task0) are being executed every minute after last valid (scheduled) execution.
3. Task01 is scheduled for 15 21 * * * * and after first run next runtime is set one hour later at 22:15:00. My time zone is CET.

Description of process:

1. The way of runnig cron tasks described in comments of TCronModule.php is not clear for me and I am running it this way:

```shell
/usr/bin/sudo -u www-data /usr/bin/php7.4 /home/roman/www/crontestapp/protected/vendor/bin/prado-cli cron/run /home/roman/www/crontestapp/
/usr/bin/sudo -u www-data /usr/bin/php7.4 /home/roman/www/crontestapp/protected/vendor/bin/prado-cli cron/tasks /home/roman/www/crontestapp/
```

2. The way of adding tasks:

```php
$cron = $this->Application->Modules['cron'];

$t = new Prado\Util\Cron\TCronMethodTask('test', 'createTmpFile3("Task0")');
$t->Name ='Task0';
$t->Schedule ='30 10 14 11 * 2022';
$t->UserName ='pracownik4';

$cron->addTask($t);

$t = new Prado\Util\Cron\TCronMethodTask('test', 'createTmpFile3("Task1")');
$t->Name ='Task01';
$t->Schedule ='15 21 * * * *';
$t->UserName ='pracownik4';

$cron->addTask($t);
```

3. My configuration from application.xml:

```xml
<modules>
    ...
    <module id="test" class="Application.Cron.TestCronModule" />
    <module id="cron_db" class="Prado\Data\TDataSourceConfig">
        <database ConnectionString="sqlite:/home/roman/www/crontestapp/protected/Files/cron.jobs" />
    </module>
    <module id="cron" class="Prado\Util\Cron\TDbCronModule" DefaultUserName="pracownik1" ConnectionID="cron_db">
        <--
        <job Schedule="55 11 * * *" Task="test->createTmpFile()" />
        <job Name="cronclean" Schedule="0 0 * * * *" Task="Prado\Util\Cron\TDbCronCleanLogTask" />
        <job Name="dbcacheclean" Schedule="0 0 * * *" Task="cache->flushCacheExpired(true)" />
        -->
    </module>
</modules>
```

4. Tests

[...]

PR #817, opened 2022-11-24
#815 Cron trigger on first run fix, repeat running fix, and removed time zone code

belisoful:
This fixes the three issues identified in #815.

1) Initializes lastExecTime for cron when loading persistent data from the global state.
1) Initializes lastExecTime for DB Cron on adding and updating (when schedule is changed).
2) Better null nextTriggerTime handling stopped the recurring tasks without a next time.
3) Also removed the Timezone code from TTimeScheduler to correct the erroneous one hour offset.

Also adds better cron shell output when there is no lastexectime (or no runs)
Adds an "@1668165071" unix timestamp format to the TTimeScheduler to reduce computation time for validation, parsing, and next trigger time of one off tasks.
TTimeScheduler regex building is cached.
Corrects a bug in the shell tableWidget column and last column wrapping.
minor documentation update to cron.

Updates the Cron unit tests.

### DIFF
--- a/framework/Shell/TShellApplication.php
+++ b/framework/Shell/TShellApplication.php
@@ -104,7 +104,7 @@ class TShellApplication extends \Prado\TApplication
 		$this->registerOption('quiet', [$this, 'setQuietMode'], 'Quiets the output to <level> [1..3], default 1', '=<level>');
 		$this->registerOptionAlias('q', 'quiet');
 
-		$this->attachEventHandler('onBeginRequest', [$this, 'processArguments'], 0);
+		$this->attachEventHandler('onInitComplete', [$this, 'processArguments'], 20);
 
 		parent::run();
 	}
--- a/framework/Shell/TShellApplication.php
+++ b/framework/Shell/TShellApplication.php
@@ -63,12 +63,12 @@ class TShellApplication extends \Prado\TApplication
 	protected $_options = [];
 
 	/**
-	 * @var array<array> application command optionAliases of the short letter(s) and option name
+	 * @var array<string, string> application command optionAliases of the short letter(s) and option name
 	 */
 	protected $_optionAliases = [];
 
 	/**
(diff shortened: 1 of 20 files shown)


belisoful:
How is Scrutinizer failing to MySql upgrade 5.5 to 5.7?   Any ideas?

ctrlaltca:
> How is Scrutinizer failing to MySql upgrade 5.5 to 5.7? Any ideas?

Fixed by updating the scrutinizer config.
To be honest tests are now run as a github action, so probably mysql/postgres are not even needed anymore on scrutinizer, but whatever.. maybe in the future we will need it again

belisoful:
The latest update did pass the scrutinizer check without any edits to the config.  It may have been some [configuration] issue from github?

Also. I think this ready for review and merge.  

I have a v2 of cron waiting in the wings to make cron friendly not just for system users but for application user use as well.  I think this bug fix merge should be logged before v2 makes the scene because v2 is an enhancement update. There are a series of micro adjustments and bug fixes to be made after cron v2 as well.

The TEmail is looking awesome.  Emailing has needed a full PRADO treatment, like Cron.  It's an exercise in what happens when a TEmail subclasses TPage.  The email module acts like a service, directives set the email properties (like the subject and priority), Emails are Localized via template to the user language, and configured like pages.  The Db Module logs emails and can send delayed emails from cron rather than directly when the page is processed.  I think a DBEmailService would be great to see, where it logs viewing of emails (by tracker image), and can display logged emails when the user is have problems in their email browser (viewed by email MessageId so only ppl with access to the email would be able to access the email by its reference Id.) Other sites are doing this now, like twitch.tv.  How can we leverage PRADO to make Emails as productive and easy as Pages?

I'm looking forward to moving past this merge.  as you can tell.

ctrlaltca (approved):
Looks fine to me

PR #830, opened 2022-12-17
Add support for OFFSET parameter without LIMIT clause in SQL query for TActiveRecord; fix #819

### LINKED ISSUE #819: Support for OFFSET parameter without LIMIT clause in SQL query for TActiveRecord
Hello the PRADO Group,

I use the PRADO framework with the PostgreSQL database. I found a small problem when I provide a SQL query with offset parameter to ``Prado\Data\ActiveRecord\TActiveRecord::findAllBySql()`` method. Here is small sample:

```
$sql = 'SELECT col FROM table ORDER BY col DESC OFFSET 20';
ObjectRecord::finder()->findAllBySql($sql);
```

I analyzed what is happening in the framework and I found that in ``Prado\Data\DataGateway\TSqlCriteria::setCondition()`` there is regex to match ``ORDER BY`` clause:

```
                 // supporting the following SELECT-syntax:
                 // [ORDER BY {col_name | expr | position}
                 //      [ASC | DESC], ...]
                 //    [LIMIT {[offset,] row_count | row_count OFFSET offset}]
                 // See: http://dev.mysql.com/doc/refman/5.0/en/select.html
 
                 if (preg_match('/ORDER\s+BY\s+(.*?)(?=LIMIT)|ORDER\s+BY\s+(.*?)$/i', $value, $matches) > 0) {
                         // condition contains ORDER BY
                         $value = str_replace($matches[0], '', $value);
                         if (strlen($matches[1]) > 0) {
                                 $this->setOrdersBy($matches[1]);
                         } elseif (strlen($matches[2]) > 0) {
                                 $this->setOrdersBy($matches[2]);
                         }
                 }
```

In case of my query this regex catches also the ``OFFSET 20`` because it matches to this regex part ``|ORDER\s+BY\s+(.*?)$``. As the result the str_replace() removes ``OFFSET 20`` and it is causes that the SQL query is without offset.

In MySQL the offset parameter we can provide in two ways:

```
LIMIT offset, row_count
```

or in this form:

```
LIMIT row_count OFFSET offset
```

and the LIMIT is obligatory for this case.

source: https://dev.mysql.com/doc/refman/8.0/en/select.html

In PostgreSQL the LIMIT is not obligatory for using offset and offset can be used without it:

```
OFFSET offset
```

source: https://www.postgresql.org/docs/current/sql-select.html

So, I would like to propose to tune a little the regex from:

```
/ORDER\s+BY\s+(.*?)(?=LIMIT)|ORDER\s+BY\s+(.*?)$/i
```

to:

```
/ORDER\s+BY\s+(.*?)(?=\s+(?:LIMIT|OFFSET))|ORDER\s+BY\s+(.*?)$/i
```

I prepared a small command line test script to check if everything is fine. You can find it here:

```
$ cat ./offset_limit_test.php 
<?php

$query_ok = [
    'SELECT col FROM table ORDER BY col DESC',
    'SELECT col FROM table ORDER BY col DESC LIMIT 10',
    'SELECT col FROM table ORDER BY col DESC OFFSET 20',
    'SELECT col FROM table ORDER BY col DESC LIMIT 10 OFFSET 20',
    'SELECT col FROM table ORDER BY col DESC LIMIT 20,10'
];

// ORIGINAL REGEX
//$test_pattern = '/ORDER\s+BY\s+(.*?)(?=LIMIT)|ORDER\s+BY\s+(.*?)$/i';

// NEW PROPOSED REGEX
$test_pattern = '/ORDER\s+BY\s+(.*?)(?=\s+(?:LIMIT|OFFSET))|ORDER\s+BY\s+(.*?)$/i';


for ($i = 0; $i < count($query_ok); $i++)
{
	echo "===> {$query_ok[$i]} <===" . PHP_EOL;
	if (preg_match($test_pattern, $query_ok[$i], $matches) === 1)
	{
		echo '[OK] ' . var_export($matches, true) . PHP_EOL;
	} else {
		echo '[ERROR]';
	}
	echo PHP_EOL;
}
```

If I run it with the original regex, the result is wrong for third query because it catches ``col DESC OFFSET 20``.

```
===> SELECT col FROM table ORDER BY col DESC <===
[OK] array (
  0 => 'ORDER BY col DESC',
  1 => '',
  2 => 'col DESC',
)

===> SELECT col FROM table ORDER BY col DESC LIMIT 10 <===
[OK] array (
  0 => 'ORDER BY col DESC ',
  1 => 'col DESC ',
)

===> SELECT col FROM table ORDER BY col DESC OFFSET 20 <===
[OK] array (
  0 => 'ORDER BY col DESC OFFSET 20',
  1 => '',
  2 => 'col DESC OFFSET 20',
)

===> SELECT col FROM table ORDER BY col DESC LIMIT 10 OFFSET 20 <===
[OK] array (
  0 => 'ORDER BY col DESC ',
  1 => 'col DESC ',
)

===> SELECT col FROM table ORDER BY col DESC LIMIT 20,10 <===
[OK] array (
  0 => 'ORDER BY col DESC ',
  1 => 'col DESC ',
)
```

If I run it with the proposed regex, everything is fine:

```
===> SELECT col FROM table ORDER BY col DESC <===
[OK] array (
  0 => 'ORDER BY col DESC',
  1 => '',
  2 => 'col DESC',
)

===> SELECT col FROM table ORDER BY col DESC LIMIT 10 <===
[OK] array (
  0 => 'ORDER BY col DESC',
  1 => 'col DESC',
)

===> SELECT col FROM table ORDER BY col DESC OFFSET 20 <===
[OK] array (
  0 => 'ORDER BY col DESC',
  1 => 'col DESC',
)

===> SELECT col FROM table ORDER BY col DESC LIMIT 10 OFFSET 20 <===
[OK] array (
  0 => 'ORDER BY col DESC',
  1 => 'col DESC',
)

===> SELECT col FROM table ORDER BY col DESC LIMIT 20,10 <===
[OK] array (
  0 => 'ORDER BY col DESC',
  1 => 'col DESC',
)
```

Also the caught one space at the end of result ``'col DESC '`` from the original regex results now are cleared ``'col DESC'``.

What do you think about this change? Thanks in advance for your opinion about it.

Best regards,
Marcin Haba (gani)

ganiuszka:
Hello The PRADO Group,

According to issue #819 I would like to add to the framework support for OFFSET parameter without LIMIT clause. In MySQL OFFSET has to be used together with LIMIT clause, but in PostgreSQL syntax OFFSET can be used without LIMIT.

Thank you in advance for accepting this change.

Best regards,
Marcin Haba (gani)

### DIFF
--- a/framework/Data/DataGateway/TSqlCriteria.php
+++ b/framework/Data/DataGateway/TSqlCriteria.php
@@ -142,7 +142,7 @@ class TSqlCriteria extends \Prado\TComponent
 		//    [LIMIT {[offset,] row_count | row_count OFFSET offset}]
 		// See: http://dev.mysql.com/doc/refman/5.0/en/select.html
 
-		if (preg_match('/ORDER\s+BY\s+(.*?)(?=LIMIT)|ORDER\s+BY\s+(.*?)$/i', $value, $matches) > 0) {
+		if (preg_match('/ORDER\s+BY\s+(.*?)(?=\s+(?:LIMIT|OFFSET))|ORDER\s+BY\s+(.*?)$/i', $value, $matches) > 0) {
 			// condition contains ORDER BY
 			$value = str_replace($matches[0], '', $value);
 			if (strlen($matches[1]) > 0) {


ctrlaltca (approved):

ctrlaltca:
Thank you!

Issue #826, opened 2022-12-09
TComponent needs a getBehaviors to return all behaviors of the class


belisoful:
something simple like this is needed:

```
	/**
	 * Returns all the behaviors attached to the TComponent.  TBehavior[s] may
	 * be attached but disabled.
	 *
	 * @return array all the behaviors attached to the TComponent
	 * @since 4.2.2
	 */
	public function getBehaviors()
	{
		return $this->_m ?? [];
	}
```

ctrlaltca:
That's fine, but i can see some problems if the returned result is modifiable by the caller.

belisoful:
Would this quell the concern?
`return $this->_m ? $this->_m->toArray() : [];`

ctrlaltca:
The array should be passed by copy (not reference), so yes

belisoful:
Holiday Bug patching cheers! This has been a long time coming.

Issue #825, opened 2022-12-09
TComponent needs a hasMethod function to check behaviors for their methods


belisoful:
It looks like this:
```
	/**
	 * Determines whether a method is defined.
	 * When enabled, this will loop through all active behaviors for the method as well.
	 * Note, event name is case-insensitive.
	 * @param string $name the method name
	 * @return bool
	 */
	public function hasMethod($name)
	{
		if (method_exists($this, $name) || strncasecmp($name, 'fx', 2) === 0 || strncasecmp($name, 'dy', 2) === 0) {
			return true;
		} elseif ($this->_m !== null && $this->_behaviorsenabled) {
			foreach ($this->_m->toArray() as $behavior) {
				if ((!($behavior instanceof IBehavior) || $behavior->getEnabled()) && method_exists($behavior, $name)) {
					return true;
				}
			}
		}
		return false;
	}
```

Test cases are written.

ctrlaltca:
Looks fine

belisoful:
There was a cron test that spontaneously failed on one of the PHP version.  Sometimes the Cron DB re-orders tasks internally.  the test needs to be corrected but isn't the biggest of concerns.

This can be closed with merge.

belisoful:
Holiday Bug patching cheers!  This has been a long time coming.

PR #837, opened 2022-12-19
Fix setter value type with prado tpl/Page

### LINKED ISSUE #836: Problem with PHP type in setter value type with Prado tpl/Page.
If we use Php type in parameter of a setter function, the expected result is not correct. Example, with a simple component : 
```
<?php

use Prado\Web\UI\ActiveControls\TActivePanel;

class TTest extends TActivePanel {
    public function setValue(bool $value): void {
        $this->setViewState("Value", $value, true);
    }
}
```

When we set the value property on a tpl  like this : 

```
<com:TTest Value="false" />
```

The value is `true` instead of `false`

zendre4:
This is for fix #836 . 

My fix ensure value for `bool`, `integer`, and `float`. 

I test this on PHP 7.4 and 8.0. If union type are used, the fix do noting.

### DIFF
--- a/framework/Web/UI/TTemplate.php
+++ b/framework/Web/UI/TTemplate.php
@@ -446,22 +446,22 @@ class TTemplate extends \Prado\TApplicationComponent implements ITemplate
 				if ($reflector = new \ReflectionClass($component)) {
 					try {
 						$params = $reflector->getMethod($setter)->getParameters();
-						if (
-							!empty($params) &&
-							isset($params[0]) &&
-							!empty($params[0]->getType()) &&
-							$params[0]->getType() instanceof \ReflectionNamedType
-						) {
-							switch ($params[0]->getType()->getName()) {
-								case 'bool':
-									$value = TPropertyValue::ensureBoolean($value);
-									break;
-								case 'int':
-									$value = TPropertyValue::ensureInteger($value);
-									break;
-								case 'float':
-									$value = TPropertyValue::ensureFloat($value);
-									break;
+						if (!empty($params) && isset($params[0]) && $type = $params[0]->getType()) {
+							if ($type instanceof \ReflectionNamedType) {
+								switch ($type->getName()) {
+									case 'bool':
+										$value = TPropertyValue::ensureBoolean($value);
+										break;
+									case 'int':
+										$value = TPropertyValue::ensureInteger($value);
+										break;
+									case 'float':
+										$value = TPropertyValue::ensureFloat($value);
+										break;
+									case 'string':
+										$value = TPropertyValue::ensureString($value);
+										break;
+								}
 							}
 						}
 					} catch (\ReflectionException $e) {
--- a/framework/Web/UI/TTemplate.php
+++ b/framework/Web/UI/TTemplate.php
@@ -11,6 +11,7 @@ namespace Prado\Web\UI;
 
 use Prado\Prado;
 use Prado\TComponent;
+use Prado\TPropertyValue;
 use Prado\Web\Javascripts\TJavaScriptLiteral;
 use Prado\Exceptions\TConfigurationException;
 use Prado\Exceptions\TException;
@@ -441,6 +442,32 @@ class TTemplate extends \Prado\TApplicationComponent implements ITemplate
 				}
 			}
 			$setter = 'set' . $name;
+			if (method_exists($component, $setter)) {
+				if ($reflector = new \ReflectionClass($component)) {
+					try {
+						$params = $reflector->getMethod($setter)->getParameters();
+						if (
+							!empty($params) &&
+							isset($params[0]) &&
+							!empty($params[0]->getType()) &&
+							$params[0]->getType() instanceof \ReflectionNamedType
+						) {
+							switch ($params[0]->getType()->getName()) {
+								case 'bool':
+									$value = TPropertyValue::ensureBoolean($value);
+									break;
+								case 'int':
+									$value = TPropertyValue::ensureInteger($value);
+									break;
+								case 'float':
+									$value = TPropertyValue::ensureFloat($value);
+									break;
+							}
+						}
+					} catch (\ReflectionException $e) {
+					}
+				}
+			}
 			$component->$setter($value);
 		}
 	}


framework/Web/UI/TTemplate.php
```diff
@@ -441,6 +442,32 @@ protected function configureProperty($component, $name, $value)
 				}
 			}
 			$setter = 'set' . $name;
+			if (method_exists($component, $setter)) {
```
belisoful, line 445: Should be updated to be behavior aware with #825

framework/Web/UI/TTemplate.php
```diff
@@ -441,6 +442,32 @@ protected function configureProperty($component, $name, $value)
...
+				if ($reflector = new \ReflectionClass($component)) {
```
belisoful, line 446: The setter/property may not be in the component but in a behavior modifying the component. looping through each behavior would be needed too but that requires #826.

framework/Web/UI/TTemplate.php
```diff
@@ -441,6 +442,32 @@ protected function configureProperty($component, $name, $value)
...
+								case 'bool':
+									$value = TPropertyValue::ensureBoolean($value);
+									break;
+								case 'int':
+									$value = TPropertyValue::ensureInteger($value);
+									break;
+								case 'float':
+									$value = TPropertyValue::ensureFloat($value);
+									break;
+							}
```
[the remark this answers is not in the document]
belisoful: setting a property value from <%= may not always result in a string, regardless of error or intention of the developer.

zendre4:
Ok @belisoful  I add the ensureString and I modify the code for stored the type in a variable, it's ok for you for this two part ? 

Regarding issue #825 what do you expect from me, I can't use hasMethod because it doesn't exist yet. I can modify my if, but it will not merge until #825 is finished...

And for #826 I don't sure to understand the change

belisoful:
Adding the ensureString is great.

When I said to store the type in a variable I mean that a "static $cache" variable needs to be created.  The cache needs to store the reflection based on the get_class($component).  Then the static cache should store the setter method type for easy and quick retrieval from the cache.  We want to only call Reflection on a class once, and we only want to get the setter type once.  Retrieving the setter type from cache is much faster than calling the same thing over and over and over.   For an example of what I am talking about, see TApplicationComponent::getClassFxEvents.  That cache the data in a static as well as relieves the data from the PRADO cache.  Both of those are things we can do to speed up this code.

#825 and #826 is urgent and that is what I am focused upon. I am hoping to have this committed in the next day or two and merged shortly there after.  You'll be able to use these methods from the main branch there after.  My humble apologies for not having these completed long ago.  

As for why #826 is important...   It is possible for a behavior to implement a setter that doesn't exist in the owner's methods.  So you'll need to loop through all the behaviors for the setters and their types beyond the main component.

Behaviors were added in 3.2.3 as a way of extending and changing the behavior of a TComponent.   Also to be up-to-date technically with Yii (in that regard).  Behaviors can add events, properties, and methods to a TComponent that it doesn't originally have.  A behavior can even make one object "look" like and act like another.

I great example I love to use for behaviors is the "student" behavior for users.  A StudentUserBehavior can add school class scheduling and grades properties to the TUser without having to have a separate structure to maintain for the user.  The TUser simply automatically gets the Student methods, events, and properties.   If you want an example, see the Permissions manager TUserPermissionsBehavior for an example of how Advanced Behaviors are used. Also see the Prado\Util\Behaviors directory.

zendre4:
ok I will wait for you to finish #825 and #826. After that, I'll use the new "hasMethod", and I'll try to use static and cache (I never used that in Prado yet)

belisoful:
Using PRADO cache may be overkill and can be implemented after the static variable is used to cache class-setter types.

What I am looking for is to only call the Reflection class once per class

ctrlaltca (changes requested):
This looks promising to me, as it solves the long time problem of interpreting the correct type for data coming from templates.
The ReflectionClass calls definitely needs some type of caching to avoid creating possibly hundreds of identical objects for a single template

framework/Web/UI/TTemplate.php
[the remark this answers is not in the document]
ctrlaltca: This seems resolved

framework/Web/UI/TTemplate.php
ctrlaltca: I guess this going to be $component->hasMethod($setter)

framework/Web/UI/TTemplate.php
ctrlaltca: Looping all the behaviors from here sounds time expensive (but can be fixed with some caching) but also a bit incorrect from an OOP point of view. 
Maybe adding a way to get the reflected method from inside TComponent itself could be an idea. This would also move the reflection cache inside the class itself, so that other callers can benefit from it.

ctrlaltca:
> The ReflectionClass calls definitely needs some type of caching to avoid creating possibly hundreds of identical objects for a single template

From a few quick tests and some literature found on the web it seems that creating ReflectionClass objects is only expensive the very first time, while following object gets created in a negligible time.
Still, if this needs to be extended to methods introduced by behaviors, we need caching for our part.

belisoful:
scrubbing data types for function parameters is also needed for setting command line data in the shell.   when `--variable=someData_1234` the `variable` is mapped to a setter handler.  The setter handler could read the function parameter data type and ensure to int, bool, float, string, array as well.

This functionality shouldn't be kept only in the template engine.

belisoful:
I think we finally have a unified enhancement for this.  

**Summary**
Setting proper type for parameters in template controls has been strange.  This has the same problem with directives, module properties, service properties, and other similar subsystems ($class, $properties, $config).

TTemplate goes through setSubProperty after the recent update.  Module Configuration has always gone through setSubProperties.  This unifies the setting of properties across the board.  I'm not sure there is much else that doesn't use setSubProperties.

The situation has gotten more complex with how parameters are multi-typed.

`TPropertyValue::ensure*` is a reasonable cut-off for types that are auto-converted.  The logic for conversion of types shouldn't be too difficult... ???

belisoful:
When #1150 is merged, this and #836 can be closed.

ctrlaltca:
Superseded by #1150

Issue #824, opened 2022-12-09
TBehaviors configuring based upon their inner config xml/php


belisoful:
IBehavior and TBehavior needs a function "init($config)" fed with the innards of their app configuration [xml tag/php array].  This is for behaviors that are more complex than just one XML tag or php properties for configuration.

belisoful:
The configuration of the behavior needs to be encapsulated in the instancing of behaviors because attaching a class behavior could be a TBehavior which would delay instancing to either attaching or instancing of the target class.

belisoful:
I'd like to wrap this up next and get it out of the way.   This is different than just a minor bug fix.  There is some engineering of behaviors involved.

What I am thinking is that the <behavior> configuration xml/array get put into its $properties under the key '=config' (IBaseBehavior::CONFIG_KEY).  Properties won't have or use the '=' because of its use in xml to set the property value.

TComponent::instanceBehavior would capture the behavior array configuration key value and unset the key. then after the behavior is instanced, it is initialized with the passed configuration data.

This follows and extends the model of using ` ['class' => 'Prado\TXClass', 'propertyA' => 'valueA']` to include its behavior configuration values.  It would look like this 
` ['class' => 'Prado\Util\Behaviors\TXBehavior', '=config' => (TXmlElement-object|config-array), 'propertyA' => 'valueA']`

This is how behaviors are initialized with configuration data:
```
	protected static function instanceBehavior($behavior)
	{
		$hasConfig = false;
		$config = null;
		$isArray = false;
		if (is_string($behavior) || (($isArray = is_array($behavior)) && isset($behavior['class']))) {
			if ($isArray && array_key_exists(IBaseBehavior::CONFIG_KEY, $behavior)) {
				$config = $behavior[IBaseBehavior::CONFIG_KEY];
				unset($behavior[IBaseBehavior::CONFIG_KEY]);
				$hasConfig = true;
			}
			$behavior = Prado::createComponent($behavior);
		}
		if (!($behavior instanceof IBaseBehavior)) {
			throw new TInvalidDataTypeException('component_not_a_behavior', get_class($behavior));
		}
		if ($hasConfig) {
			$behavior->init($config);
		}
		return $behavior;
	}
```

Also, IBaseBehavior adds a new abstract method init($config).  TClassBehavior needs that init method, in kind.

[...]

PR #842, opened 2022-12-21
#824 Fixed: behaviors are initiated with configuration data

belisoful:
with test unit.

### DIFF
--- a/framework/Util/IBaseBehavior.php
+++ b/framework/Util/IBaseBehavior.php
@@ -23,6 +23,16 @@ namespace Prado\Util;
 interface IBaseBehavior
 {
 	/**
+	 * The array key for the $config data in instancing behaviors with init($config),
+	 */
+	public const CONFIG_KEY = "=config";
+	/**
+	 * Handles behavior configurations from TBehaviorsModule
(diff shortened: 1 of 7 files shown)


framework/Util/IBaseBehavior.php
```diff
@@ -22,6 +22,16 @@
  */
 interface IBaseBehavior
 {
+	/**
+	 * The array key for the $config data in instancing behaviors with init($config),
+	 */
+	public const CONFIG_KEY = "=config";
```
ctrlaltca, line 28: Is "=config" intended to start with "="? To avoid people from setting it accidentally?

belisoful: exactly.

belisoful: '=' isn't a supported character for tags in xml.

ctrlaltca (approved):

belisoful:
BTW, this is an important enhancement.  the new TAsset publishing code intends to do meta tag filtering and setting and setting new meta tags in jpegs would be be served not as just a class property but as a behavior xml tags on their own.

Issue #838, opened 2022-12-20
IClassBehavior should have an Enabled property like IBehavior


belisoful:
The get/set Enabled needs to be moved from IBehavior to IBaseBehavior so that IClassBehavior also gets the Enabled property.

TComponent Needs to key on IBaseBehavior for the getEnabled method rather than IBehavior.

There should probably be a TBaseBehavior class that contains the core functions.

belisoful:
This issue requires much better abstraction in the behaviors/class-behaviors.  This will also basically resolve #891 and #851 in the rewrite.  Things will basically stay the same or move up an abstraction, so there should be no issue with compatibility.

The unit tests need to be rejiggered for the new behavior functionality.

PR #849, opened 2022-12-24
Fixes #847 Adds IPriorityProperty & updates TPriority* to set priority on items

### LINKED ISSUE #847: Collections:IPriorityItem needs a subclass for setting the Priority when an item is added
I propose that a new class IPriorityProperty be created for calling setPriority when an item is added.  This way the item can capture the priority it is being added.

Both TPriorityList and TPriorityMap would need to setPriority when the item is an IPriorityProperty and the priority is different from the IPriorityProperty getProperty.

#838 gives Priority properties to the IBaseBehaviors.  This way behaviors can know their priority.

belisoful:
This is needed for bug fixes to Behaviors in #838

This also abstracts TPriority*::ensurePriority for setting default priority to null values, rounding to precision, and converting to string.

Updates TWeakCallableCollection for IPriorityProperty as well.

With unit tests. much better unit tests for TPriorityMap regarding priorities.

### DIFF
--- a/framework/Collections/IPriorityProperty.php
+++ b/framework/Collections/IPriorityProperty.php
@@ -1,0 +1,28 @@
+<?php
+/**
+ * IPriorityProperty interface file
+ *
+ * @author Brad Anderson <belisoful@icloud.com>
+ * @link https://github.com/pradosoft/prado
+ * @license https://github.com/pradosoft/prado/blob/master/LICENSE
+ */
+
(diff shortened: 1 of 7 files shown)


ctrlaltca (approved):

belisoful:
This is a huge update in a small package.  Behaviors can insert their events at their priority now that they can know their attached priority.

Issue #856, opened 2022-12-31
Intermittent TSimpleDateFormatterTest failures (on last day of the month that has 31 days)


belisoful:
I am getting these exceptions during testing without change to the code.  I think there is some time dependency to this bug...  I've gotten it before and it went away on its own.  I think this bug occurs on the last day of the month.

Thinking about this...  "09" is trying to be interpreted as the current day, but there is no 31st day in September...  September only goes to 30 days.

This brings up the question of what to do when the current day is beyond the number of days in the month being computed.

I suggest the behavior is, in some way, correct.  The unit test should be changed to be on a month that has 31 days rather than 30.  This would disrupt commits on the last day of about every other month [that has 31 days].  strange bug.

```
1) TSimpleDateFormatterTest::testParseMonthPattern
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'2022-09-31'
+'2022-10-01'

/Users/user/Code/2022prado/protected/vendor/pradosoft/prado/tests/unit/Util/TSimpleDateFormatterTest.php:127
/Users/user/Code/2022prado/protected/vendor/pradosoft/prado/vendor/bin/phpunit:123

2) TSimpleDateFormatterTest::testParseDayPattern
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'2019-09-31'
+'2019-10-01'

/Users/user/Code/2022prado/protected/vendor/pradosoft/prado/tests/unit/Util/TSimpleDateFormatterTest.php:174
/Users/user/Code/2022prado/protected/vendor/pradosoft/prado/vendor/bin/phpunit:123
```

belisoful:
Thinking about it again....  This is not the right behavior.  parse should increment the month.

A unit test on this can be written.

belisoful:
I suggest updating the unit test.  127, 130, 131 should be the October (10), not September (9).

A separate unit test should check for "Y-09-31" that it be "Y-10-01"

ctrlaltca:
fixed in #857

PR #857, opened 2022-12-31
#856 corrected TSimpleDateFormatterTest unit, the 09 month => 8

belisoful:
This is a very time sensitive change request.    any commits to any branches today will fail without this.  Only on months with 31 days, on the last day will this unit test always fail.


belisoful:
This is only changing the unit tests. The number of days in the month of September, being only 30, was the problem with the test unit.  The erroneous 9-31 is also tested as 10-01

ctrlaltca (approved):

PR #860, opened 2023-01-01
#847 just IPriorityCapture changes

belisoful:
Added IPriorityCapture to capture changes to priority but not set them.  IPriorityProperties is the interface for extending both giving and receive priority.

### DIFF
--- a/framework/Collections/IPriorityCapture.php
+++ b/framework/Collections/IPriorityCapture.php
@@ -1,0 +1,27 @@
+<?php
+/**
+ * IPriorityCapture interface file
+ *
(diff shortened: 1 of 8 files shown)


belisoful:
my apologies that the last pull request that had the erroneous change to the composer file.  That's been corrected.

belisoful:
Sorry about the last IPriorityCapture PR having the change to the composer file.  that was an accident from being very distracted here.

belisoful:
Ready for review.

ctrlaltca (approved):

PR #867, opened 2023-02-23
Define the html purifier cache in the runtime directory

### LINKED ISSUE #866: HTMLPurifier_Config and cache path
Hi,

Prado is using HTMLPurifier for sanitization TTextBox and TSafeHtml. HTMLPurifier use a internal cache where the default path, if you don't configure one, is in its own library directory. Meaning in the vendor folder. Of course, vendor should not be writable for security reason. I propose to set the default cache to the runtime path like this for the TTextBox:

```
public function getConfig()
{
	$config = $this->getViewState('Config', null);
        if ($config === null) {
            $config = \HTMLPurifier_Config::createDefault();
            $config->set('Cache.SerializerPath', Prado::getApplication()->getRuntimePath());
        }
	return $config;
}
```
The same thing could be done for TSafeHtml.

If you are agree, I can propose a PR.

majuca:
resolve #866


ctrlaltca:
Looks good, but CI is complaining because the code is using spaces instead of tabs for indentation.

majuca:
ok, I see, I will fix it.

ctrlaltca:
Merged, thank you!

PR #871, opened 2023-02-28
Implement THttpRequest::ResolveMethod

### LINKED ISSUE #868: RFC: Service detection is fragile
An PRADO application can expose multiple services; usually there's a single "page "service, but more can be defined in application.xml.
The way PRADO detects which service to run is implemented in THttpRequest: https://github.com/pradosoft/prado/blob/master/framework/Web/THttpRequest.php#L761
This looks a bit fragile.

I have an example application containing 2 services: a "page" service hosting a website and a "geonode" service, accessed by desktop applications like [QGIS](https://www.qgis.org/it/site/).
Requests to the "page" service runs fine with urls like http://127.0.0.1/prado-app/page,Home.
Requests to the "geonode" service are a bit more complex and contains parameters hardcoded in the client, eg: http://127.0.0.1/prado-app/geonode/api/v2/layers/?page=1&page_size=10&sort[]=name&filter%7BstoreType.in%7D=dataStore&filter%7BstoreType.in%7D=coverageStore
A problem arises on the second request: Prado merges all the request parameters in a single array, and loops through the array to look for a service id. The "page=1" parameter gets interpreted as a request to the "page" service, and since it's the first service matching in the array it gets used, even if it's not the correct one.
Now, since the request parameters are hardcoded in the client (https://github.com/GeoNode/QGISGeoNodePlugin/blob/main/src/qgis_geonode/apiclient/geonode_v3.py#L56), i can't really fix them but i need a way to workaround the problem, forcing PRADO to use the correct service.

I'll try to workaound this in my app overloading the THttpRequest::resolveRequest() method, hopefully finding a way extract all matching services and giving priority to the one appearing first.
If you have any comment on other ways to fix this, i'd be glad to hear from you :)

ctrlaltca:
This new parameter instructs THttpRequest on the behavior used to determine the correct service to be used to handle the request
Fix #868

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -2,6 +2,8 @@
 ## Version 4.2.2 - ____
 
 ENH: Issue #828 - TApplication::onSetUser raised when setting user. (belisoful)
+ENH: Issue #866 - HTMLPurifier_Config and cache path (majuca)
+ENH: Issue #868 - RFC: Service detection is fragile (ctrlaltca)
 BUG: Issue #815 - Cron tasks delay until their proper time on first entry. Long running cron tasks no longer repeat further pending tasks. (belisoful)
 
 ## Version 4.2.1 - May 9, 2022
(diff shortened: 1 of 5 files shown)


majuca:
For me it is ok, should I merge it?

majuca (approved):
Ok for me!

PR #873, opened 2023-03-02
Defaut permission for files and directories

### LINKED ISSUE #865: Files permissions created by Prado
Hi,

Is there any particular reason to have set the default permissions to 0777 for the PRADO_CHMOD constant? I think that for security reasons it would be better to have the value at 0755.

Shouldn't there be two constants, one for files at 0644 (PRADO_FILE_CHMOD) and one for directories (PRADO_DIR_CHMOD) at 0755? There is no reason to put execution rights on files.

Jean-Luc

majuca:
Two new functions was added as proposed by @ctrlaltca Prado::getDefaultDirPermissions() and Prado::getDefaultFilePermissions(). The function Prado::getDefaultPermissions is now deprecated. 

The const PRADO_CHMOD is also deprecated and must be replace (if required) by PRADO_DIR_CHMOD and PRADO_FILE_CHMOD. For backward compatibility, until the next major release, the define PRADO_CHMOD has priority if defined in the index.php.

Fix #865

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -3,7 +3,7 @@
 
 ENH: Issue #828 - TApplication::onSetUser raised when setting user. (belisoful)
 BUG: Issue #815 - Cron tasks delay until their proper time on first entry. Long running cron tasks no longer repeat further pending tasks. (belisoful)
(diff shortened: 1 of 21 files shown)


ctrlaltca (approved):
Looks good to me

PR #876, opened 2023-03-15
Make Prado::using able to autoload traits defined as prado3 namespaces

### LINKED ISSUE #875: Use of Php traits
I'm using Prado 4.2.1 and I can't use Php Traits.
Is that normal? How can we use traits with Prado?

ctrlaltca:
Fix #875


belisoful:
Nice.

PR #878, opened 2023-04-06
Php updates, prepare release 4.2.2

ctrlaltca:


belisoful:
What is up with the trailing comma on all arrays (with their listed elements)?

Is that a feature of PHP 8 to vaporize trailing empty array elements?

ctrlaltca:
Those changes were introduced by php-cs-fixer. Trailing commas are now allowed both in arrays, function calls and function definitions. I guess always adding a comma is considered more error proof?

PR #879, opened 2023-04-06
Php8 only support

ctrlaltca:
This PR has been opened to show the current progress on requiring php 8 (8.1, actually) and porting code to the use of php8 -only features.

### DIFF
--- a/.github/workflows/prado.yml
+++ b/.github/workflows/prado.yml
@@ -26,7 +26,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-versions: ['8.0', '8.1', '8.2']
+        php-versions: ['8.1', '8.2']
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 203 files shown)


ctrlaltca:
This PR is already huge enough by now to be considered to be merged.
The minimum required pp version has been bumped to 8.1.
The build/test system has been updated accordingly, and some checks in the code to support previous php versions have been removed.
All the class names hardcoded in strings have been replaced by equivalent MyClass::class calls, that are more useful for static analysis of the code. Calls to get_class($object) have been replaced by $object::class, too.
Some other php-specific changes:
 * json_encode/decode() are finally able to throw exceptions, so we don't need to capture their errors and convert them to exceptions anymore;
 * TFirebugLogRoute was broken and has been fixed
 * TTextBox had support for some autocomplete types based on vcard dating back to internet explorer 4, that have never been supported by anything else and are now removed.

Issue #861, opened 2023-01-12
TPropertyValue::ensureHexColor to ensure a web color '#RRGGBB'


belisoful:
In working on the TAsset publishing sub-system, there is a some color processing.  Colors need to be specified.

I think it would be great to have a TPropertyValue::ensureHexColor that ensures web colors and web hex colors.

Here is what I propose.  Prado\Web\UI\TWebColor Enumerable with all the web color names and their # hex color value.

The ensureHexColor would index TWebColor to translate web colors to their hex colors.  It would ensure that #RGB is converted into #RRGGBB.  and nothing but a web hex color will validate.   A numeric color value (0xAARRGGBB) would be translated into a web hex color.  (no alpha, obviously)

I suggest that 3 parameters be provided.  When the 3rd parameter is provided, ensureHexColor will accept a red, green, and blue color value [0..255] and translate that into a web hex color.

This will be part of the updated publishing sub-system.  I wanted to put it into an Issue for commentary and documentation.

I already have it coded, working and have the unit tests written.  It's looking really great.  It's a good addition to the "ensure" functionality.

belisoful:
This is looking easy and is working right.  It just needs to be isolated and uploaded.

belisoful:
excellent.   We've been needing this standardized list of Web Colors and associated ensure function.  The unit test for TPropertyValue has been sorely needed too.  I'm a fan of the multi-utility of the ensureHexColor.

belisoful:
One more edit before completion.

Checking the $value for is_array.  and set the $blue, $green, and $value to their $value['blue', 'green', 'red'] or [2, 1, 0], and then proceed with the rest of the method.

belisoful:
I am very happy with this merge.  I like all the web colors finally being available in a list in PRADO.  I like the ensureHexColor in TPropertyValue.  the new #835 Asset publishing uses color and color processing in configuration.

PR #881, opened 2023-04-12
#861 TWebColors, ensureHexColor, TPropertyValue unit tests

belisoful:
This adds all the web colors into a TEnumerable.
TPropertyValue::ensureHexColor uses the new TWebColors to convert colors into their hex color. We finally have basic unit tests for TPropertyValue.  While this may seem redundant, It provides very clear cases on how it works and is to be used.

It also checks the Web Colors for proper work.


ctrlaltca (approved):
Looks good to me. Thank you for all the TPropertyValue tests!

Issue #843, opened 2022-12-21
Permissions Behaviors::"getManager" should be more specific like "getPermissionsManager"


belisoful:
This affects TPermissionsConfigurationBehavior, TPermissionsBehavior, and TUserPermissionsBehavior.

This avoids collision with any other Behaviors and their Managers.

PR #883, opened 2023-04-12
#843 changed PermissionsBehaviors::getManager to getPermissionsManager for specificity

belisoful:
basically, as behaviors, these shouldn't hog the "getManager" name from other behaviors, and it needed to be more specific on its own.


ctrlaltca:
Ping me when this is this ready

belisoful:
There aren't any further issues on this.  it can be merged.

ctrlaltca (approved):

Issue #845, opened 2022-12-22
PHP Clone and Unserialize of TComponent objects support for behaviors


belisoful:
TComponent needs a __clone magic method that loops through _m, clones IBehaviors, and re-attaches all the behaviors to the cloned object,  in order.  Also, events need to be cloned and likely scrubbed of behavior events.

I've done some research into PHP and what is going on when cloning.  The clone __construct is called, where it gets class behaviors.  All the set values and preloaded behaviors are then clobbered by a shallow copy.   Then __clone is called on the cloned object with the shallow copied data.

Anything that the __construct class behavior does to the TComponent is basically undone with the clone shallow copy.  IBehaviors should be cloned and then attached to the new cloned owner with proper retained data.  IClassBehaviors are just passed through and attached to the new Owner, after setting proper retained data.

There is going to be erroneous Class Behaviors (IBehaviors and IClassBehaviors) attached which get vaporized.  Those have weak references in various attached app/object 'on' or 'fx' events that are no issue.  The WeakReferences should allow the erroneous behavior to vanish and the handlers become void.

These new (suggested, incomplete, uncommitted) unit tests in TComponentTest are failing:
```
	public function testClone()
	{
		$this->component->attachBehavior('CopyBehavior', $b = new FooFooBehavior());
		
		$this->assertNotNull($this->component->asa('CopyBehavior'));
		$this->assertEquals($this->component, $this->component->CopyBehavior->getOwner());
		
		$copy = clone $this->component;
		$copy->Text = 'copyObject';
		
		$this->assertNotNull($copy->asa('CopyBehavior'));
		$this->assertEquals($this->component, $this->component->CopyBehavior->getOwner());
		$this->assertEquals($copy, $copy->CopyBehavior->getOwner());
		$this->assertFalse($copy->CopyBehavior === $this->component->CopyBehavior);
	}
```

This unit test is not enough to test this thoroughly yet, but it is a start of what I'm thinking the result should be.

Here is the function in TComponent for passing these tests:

```
	/**
	 * This reattaches the behaviors to the cloned object.
	 * IBehavior objects are cloned, IClassBehaviors are not.
	 * Events are scrubbed of behaviors' events.
	 * @since 4.2.2
	 */
	public function __clone()
	{
		
		foreach ($this->_e as $event => $handlers) {
			$this->_e[$event] = clone $handlers;
		}
		$behaviors = array_values(($this->_m !== null) ? $this->_m->toArray() : []);
[...]

PR #884, opened 2023-04-14
#845 Adds TComponent::__clone, __wakeup, and callBehaviorsMethod

belisoful:
TComponent: adds __clone, __wakeup, and callBehaviorsMethod.
     * __clone removes former behaviors' event handlers and attaches the cloned behaviors to the new object.  At the end, dyClone is raised for attached behaviors to finalize themselves.
     * __wakeup attaches unserialized behaviors and uses global behaviors over old instances' unserialized copy, then attaches global behaviors when not found in the object. At the end, dyWakeUp is raised for attached behaviors to finalize themselves.
     * callBehaviorsMethod is to call a method or raise a dynamic event on a TComponent's behaviors.
-TBehavior: new __clone method resets the owner, adds _getZappableSleepProps to not save the owner.  Owners are set on unserialize when attaching.
- Adds parent::__clone to existing classes __clone method [TStyle].  It should typically be called after a class's own initialization

Updated unit tests unit tests, added tests for __clone and __wakeup.   callBehaviorsMethod are covered by the existing unit tests for __call.

There are a few updates the syntax of function calls.

Some TList and TMap have better Zapping sleep Props and are more strictly type cast with the upgrade to PHP 8.  The better serialization in TComponent uses the serialization of the list/map...  so the update kinda goes together.

A minor bug in error display in TApplication is corrected as well.

### DIFF
--- a/framework/Collections/TAttributeCollection.php
+++ b/framework/Collections/TAttributeCollection.php
@@ -137,7 +137,7 @@ class TAttributeCollection extends TMap
 	 * @param mixed $key the key
 	 * @return bool whether the map contains an item with the specified key
 	 */
-	public function contains($key)
+	public function contains($key): bool
 	{
 		return parent::contains($this->_caseSensitive ? $key : strtolower($key));
 	}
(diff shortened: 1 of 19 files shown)


belisoful:
If this is too much, I can split the Collections typecasting into a separate merge so this is just the core of #845.

All these changes are split into separate commits.

Otherwise. I've already spent much time going over #845 again and again and again to get it as near perfect as possible.

On my side, this is ready to merge.

belisoful:
FYI.  The bug fix for TWeakCallableCollection uses some of the TList and TMap updates here.

PR #887, opened 2023-04-15
#861 Added EnsureHexColor support for array type colors

belisoful:
This adds support for an array of color [$r, $g, $b] or ['red' => $r, 'green' => $g, 'blue' => $b].  eg. a color can be retrieved by GD and converted into an RGB Hex Color.

Proper documentation and unit tests included.


belisoful:
Normally, ensureXXXX. only takes string... and ensureHexColor does that.  but I'd like it to be a multifunction for producing string hex colors from a variety of sources.  One source is GD that produces an array of ['red' => 255, 'green' => 255, 'blue' => 255].

This merge accounts for these array colors from GD in producing a hexicolor.

ctrlaltca (approved):

PR #899, opened 2023-04-20
TPriorityList and TPriorityMap updates for less variable copying and better organization

belisoful:
This adds the TPriorityCollectionTrait to simply the TPriorityList/Map classes.

TPriorityList is made more efficient in reducing variable/array copying and has better error checking.  This also adds support for copyFrom and MergeWith of TPriorityMap; where the keys are dropped and put in the order given.

TPriorityMap is made more efficient in reducing variable/array copying and add support for copyFrom and MergeWith a TPriorityList where the list index is treated as the key in the map.  List index numbers are the new keys at the specified priority.

TWeakCallableCollection gets two new functions getIterator and toArray to account for changes in the TPriorityList code.

The new traits and classes are added the classes.php list.

Unit tests for the new code.

### DIFF
--- a/framework/Collections/TPriorityCollectionTrait.php
+++ b/framework/Collections/TPriorityCollectionTrait.php
@@ -146,15 +146,14 @@ trait TPriorityCollectionTrait
 
 	/**
 	 * This flattens the priority list into a flat array [0,...,n-1]
-	 * @return array array of items in the list in priority and index order
 	 */
-	protected function flattenPriorities(): array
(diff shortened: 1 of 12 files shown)


belisoful:
@ctrlaltca Can you do a brief performance profile on this merge and compare it to the prior code?  I'm interested in seeing what kind of performance boost you are getting from this merge.   Lists should be more performant with this update.

belisoful:
Ready for Merge

Issue #864, opened 2023-02-09
TUtf8Converter needs language settings for iconv and TEscCharsetConverter for converting ESC charset and iconv charset


belisoful:
Namaste!

This is a request for information regarding personal experience, info, webpages, algorithms, and/or code with implementing/parsing Escape Character Code Sets for conversion into iConv::charEncoding.

In IPTC, strings in the second record set and above are stored in a character encoding specified (possibly) in the IPTC.  These character encodings are stored in escape data.  That escape data needs to be read and parse.   The IPTC char set is then used to convert all proper string data to UTF-8 internally.

I have a halfway decent first iteration of the IPTC CharSet Escape Char Code into iConv::charEncoding.

Esc control characters (controlling the char set) is found here:
https://en.wikipedia.org/wiki/ISO/IEC_2022

and ANSI and the national variants with the ESC values are here:
https://en.wikipedia.org/wiki/ISO/IEC_646

The thing I'm looking for experience and commentary on the ESC char set conversion.  Is there a PHP method I am missing to make this easier?   Is there an easy way to process ESC character control characters?

How would you parse ESC char set data in IPTC?

Here is the official IPTC spec:
http://www.iptc.org/std/IIM/4.1/specification/IIMV4.1.pdf

And the Character Encoding is found in '1#90'. record 1, tag id 90.  Common Character encodings are in appendix C.

How is G0, G1, G2, and G3 handled differently?  How is C0 and C1 handled?  In reading the ESC char set, do the other ESC values need to be restored if the charset changes?

My current thinking is to have a configurable default IPTC Charset when an IPTC has no charset.  Then have an Output IPTC charset configuration (None, Preserve, etc ['UTF-8', 'CP1252',...]).  The default input is ASCII and the default output is UTF-8.   This is the default behavior of other platforms as well.

belisoful:
The way iconv accounts for national variations of ASCII is with "setlocale".   At least that is what I have gathered so far.

Can someone confirm?

ctrlaltca:
In the documentation of the function iptcparse: https://www.php.net/manual/en/function.iptcparse.php i've foud this snippet of code that may be relevant:
```php
$IPTC_Caption = "";
$size = getimagesize( $image_path, $info );
if (isset($info["APP13"])) {
    if($iptc = iptcparse( $info["APP13"] ) ) {
        $IPTC_Caption = str_replace( "\000", "", $iptc["2#120"][0] );
[...]

PR #901, opened 2023-04-20
#864 TUtfConverter takes language for national localization and TEscCharsetConversion for converting ESC charsets

belisoful:
```
toUTF8($string, $from, $lang = null)
fromUTF8($string, $to, $lang = null)
```
these functions add the $lang parameter for setting the PHP setLocale(LC_CTYPE, $lang) because various countries/languages have slightly different character sets despite being the same encoding.  eg ASCII has different national standards.

This is the most comprehensive list of ESC character set encodings i was able to find in reasonable time.

### DIFF
--- a/framework/Util/TEscCharsetConverter.php
+++ b/framework/Util/TEscCharsetConverter.php
@@ -101,10 +101,11 @@ class TEscCharsetConverter
 	 * @return ?string The ESC character code representing the encoding
 	 *   or null if not found.
 	 */
-	public static function encodeEscapeCharset(string $charset): string
+	public static function encodeEscapeCharset(string $charset): ?string
 	{
-		if(($escEncoding = array_search($charset, self::ESC_CHAR_ENCODINGS_MAP) !== false)
-			return $charset;
+		if (($escEncoding = array_search($charset, self::ESC_CHAR_ENCODINGS_MAP)) !== false) {
+			return $escEncoding;
+		}
 		return null;
 	}
 }
--- a/framework/Util/TEscCharsetConverter.php
+++ b/framework/Util/TEscCharsetConverter.php
@@ -74,7 +74,7 @@ class TEscCharsetConverter
 			"\x1B\x24\x28\x4F" => 'ISO-2022-JP-3',	"\x1B\x24\x29\x4F" => 'ISO-2022-JP-3',	"\x1B\x24\x2A\x4F" => 'ISO-2022-JP-3',	"\x1B\x24\x2B\x4F" => 'ISO-2022-JP-3',
 			"\x1B\x24\x28\x50" => 'ISO-2022-JP-3',	"\x1B\x24\x29\x50" => 'ISO-2022-JP-3',	"\x1B\x24\x2A\x50" => 'ISO-2022-JP-3',	"\x1B\x24\x2B\x50" => 'ISO-2022-JP-3',
 		];
-	
+
 	/**
 	 * Convert an Escape Character Code Encoding to the iconv character
 	 * encoding.
--- a/framework/Util/TEscCharsetConverter.php
+++ b/framework/Util/TEscCharsetConverter.php
@@ -1,0 +1,110 @@
+<?php
+/**
+ * TEscCharsetConverter class file
+ *
+ * @author Brad Anderson <belisoful@icloud.com>
+ * @link https://github.com/pradosoft/prado
+ * @license https://github.com/pradosoft/prado/blob/master/LICENSE
+ */
+
+namespace Prado\Util;
+
+/**
+ * TEscCharsetConverter class.
+ *
+ * TEscCharsetConverter is the ESC Charset Converter for converting between ESC
+ * character sets] encodings and their iConv character encodings.
+ *
+ * Each Esc charset Encoding has 4 versions for G0, G1, G2, and G3.
+ *
+ * @author Brad Anderson <belisoful@icloud.com>
+ * @since 4.2.3
+ * @see https://en.wikipedia.org/wiki/ISO/IEC_2022 General structure of character encodings.
+ * @see https://en.wikipedia.org/wiki/ISO/IEC_646 National standards for ASCII.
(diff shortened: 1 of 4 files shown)


framework/Util/TUtf8Converter.php
```diff
@@ -17,20 +17,81 @@
...
 	 * @return string UTF-8 encoded string, original string if iconv failed.
 	 */
-	public static function toUTF8($string, $from)
+	public static function toUTF8($string, $from, $lang = null)
 	{
 		if ($from != 'UTF-8') {
+			$locale = null;
+			if ($lang !== null) {
+				$locale = setLocale(LC_CTYPE, '0');
+				setLocale(LC_CTYPE, $lang);
```
ctrlaltca, line 89: I hate the way PHP handles this.
Locale on PHP is per-process, so any other thread executing at this point would inherit this, even if it's a complete separate app.
Still, it's the only way to do it in PHP, so it's ok

framework/Util/TUtf8Converter.php
```diff
@@ -17,20 +17,81 @@
...
+ * These are not yet in iconv (as of April, 2023):
+ * @todo missing ISO-5427. https://en.wikipedia.org/wiki/ISO_5427. (8 bit Cyrillic, 1979/1981)
+ * @todo missing ČSN (Czech technical standard) 369103. https://en.wikipedia.org/wiki/KOI_character_encodings (also Cyrillic)
  */
 class TUtf8Converter
 {
+	public const ESC_CHARENCODINGS_MAP = [
+			"\x1B\x25\x47" => 'UTF-8', // ESC-'%G'
+
+			"\x1B\x28\x40" => 'ASCII',	"\x1B\x29\x40" => 'ASCII',	"\x1B\x2A\x40" => 'ASCII',	"\x1B\x2B\x40" => 'ASCII',
```
ctrlaltca, line 33: I'm not sure this is the best class to put this array of "escape used to represent encodings", since it's outside the scope of UTF8 itself.
Still, it's the most matching class actually existing in prado, so unless you want to create another helper class, it's ok if they stay here by now.
Out of curiosity: it looks like each encoding has multiple escape codes, but i've not been able to understand why.

ctrlaltca (approved):

framework/Util/TUtf8Converter.php
belisoful: I know.  When I encountered this I thought the same thing.  PHP is doing the dirty on this function.

framework/Util/TUtf8Converter.php
belisoful: I was debating if this is the right place to put it.  I think you may be right about this being in its own TEscCharsetConverter.

As for why there are 4 versions of each...  It has to do with G0, G1, G2, and G3 versions of each.  I do not know why there are G0-G3 versions or what each version does.   As far as I have been able to tell, they map the same way, but my experience is limited despite looking these things up.

belisoful:
While I was sleeping on this PR, it came to me that if $lang is null, to check the $from/$to for a '.', and then pull the encoding and lang apart.

framework/Util/TUtf8Converter.php
belisoful: This is also why the length of time in setting the Locale is minimized: to avoid collisions.  The mapping of character sets isn't "highly" influenced by Locale, there are only a few minor differences like '$' mapping to "£" in England and "€" in Europe.

This is why the $lang setting is null by default and optional, and typically will not be used.  Nothing existing will use it and will avoid PHP collisions of Locale in the per-process.

That said:
IPTC character encoding has the requirement of its own separate charset encoding language apart from the web user.

An image could be encoded anywhere and the Web User not be located there.

belisoful: The language change on the process is limited to LC_CTYPE conversion.

framework/Util/TEscCharsetConverter.php
```diff
@@ -0,0 +1,111 @@
...
+ * @see https://www.sljfaq.org/afaq/encodings.html Japanese encodings and character sets.
+ *
+ * These are not yet in iconv (as of April, 2023):
+ * @todo missing ISO-5427. https://en.wikipedia.org/wiki/ISO_5427. (8 bit Cyrillic, 1979/1981)
+ * @todo missing ČSN (Czech technical standard) 369103. https://en.wikipedia.org/wiki/KOI_character_encodings (also Cyrillic)
+ */
+class TEscCharsetConverter
+{
+	public const ESC_CHAR_ENCODINGS_MAP = [
+			"\x1B\x25\x47" => 'UTF-8', // ESC-'%G'
```
belisoful, line 33: Just wanted to add that these are static and set for the long term.  They haven't changed since they were initially set.  Missing and new should be added on the rarest of times.

There are dates beside the @ todo so we know how long it has been since updated/checked.

belisoful:
I have a IPTC class for reading and writing IPTC that will be using this updated TUtfConverter and Esc charset converter at some point..

framework/Util/TUtf8Converter.php
ctrlaltca: Now i get it, thank you

ctrlaltca:
> While I was sleeping on this PR, it came to me that if $lang is null, to check the $from/$to for a '.', and then pull the encoding and lang apart.

Looking at the output of `iconv -l` i can see that some charset already contains a dot in their name.
It looks like they are all quite exotic:

ANSI_X3.4-1968, ANSI_X3.4-1986, ANSI_X3.4, ANSI_X3.110-1983, ANSI_X3.110
CSA_Z243.4-1985-1, CSA_Z243.4-1985-2, CSA_Z243.419851, CSA_Z243.419852
ISO_646.IRV:1991, JUS_I.B1.002, MSZ_7795.3, T.61-8BIT, T.61, T.618BIT,
TIS620.2529-1, TIS620.2533-0
I guess we can like without these.. LGTM

ctrlaltca (approved):

Issue #827, opened 2022-12-09
TTemplateManager needs  instancing subclasses of TTemplate and input culture for non-user based localization


belisoful:
I'm working on adding a TEmailTemplate for capturing and dealing with published files as: links, automatic email attachments, or embedded data.  It acts as a publish rerouter.  Default behavior is like a webpage, the asset is publish in a the assets folder and linked in the email.

However, TTemplateManage need to be able to accept a subclass of TTemplate, and culture for localizations not based upon the request/session.  A.TEmailTemplate could be sent to a user where the active web user is not the email receiver and needs localization for that specific user.

eg. mail list module where users languages are captured when they subscribe and used to customize future emails to them.

Such templates would get the caching speed bump too.

ctrlaltca:
A property could be added to TTemplateManager that can be set to the class to be used for generated templates, eg. what is currently used by TUrlMapping for DefaultMappingClass:
```php
class TTemplateManager ...
...
	protected $_defaultTemplateClass = '\Prado\Web\UI\TTemplate';

	public function getDefaultTemplateClass()
	{
		return $this->_defaultTemplateClass;
	}

	public function setDefaultTemplateClass($value)
	{
		$this->_defaultTemplateClass = $value;
	}
```

TTemplateManager::getTemplateByFileName() would them make used of this class name to instantiate new templates.

Localization is a bit harder and would probably require a few more methods.

belisoful:
The surprisingly simple solution i have working right now is to redefine this:
https://github.com/pradosoft/prado/blob/fcc6b66a5aea385f163335e18f2e28d22cdd3c6c/framework/Web/UI/TTemplateManager.php#L75

to this
`public function getTemplateByFileName($fileName, $tplClass = "\\Prado\\Web\\UI\\TTemplate", $culture = null)`

First to check that the $tplClass is an instance of TTemplate for sanity. Then instance the class from that.  $culture is passed through to the localize function. null just uses the application globalization.

It's not really a change of the defaultTemplateClass,  It's an overriding.  The TEmailerModule doesn't need to change the default, it should work in parallel to the default.  Working on the TEmail functionality is providing a lot of unique insight into how PRADO can work better.

ctrlaltca:
Simple solutions are always the best.
Still, adding a DefaultTemplateClass property would make this configurable in application.xml

belisoful:
I see what you are suggesting with the DefaultTemplateClass.  So i agree.  That does make sense.

[...]

PR #902, opened 2023-04-20
#827 TTemplateManager accepts template class and culture

belisoful:
With the option of setting a new default template class upon instancing of TTemplateManager.

This change allows a template to be formed upon a designated culture rather than just the web user's culture settings.  this is important for when sending emails to someone with different culture settings.  The template must be formed around the email receiver and not the web user.

### DIFF
--- a/framework/Web/UI/TTemplateManager.php
+++ b/framework/Web/UI/TTemplateManager.php
@@ -11,6 +11,7 @@ namespace Prado\Web\UI;
 
 use Prado\Prado;
 use Prado\TApplicationMode;
+use Prado\TPropertyValue;
 
 /**
  * TTemplateManager class
@@ -42,6 +43,10 @@ class TTemplateManager extends \Prado\TModule
 	 * Prefix of the cache variable name for storing parsed templates
 	 */
 	public const TEMPLATE_CACHE_PREFIX = 'prado:template:';
+	/**
+	 * @var string Default template class, default '\Prado\Web\UI\TTemplate'
+	 */
+	private string $_defaultTemplateClass = TTemplate::class;
 
 	/**
 	 * Initializes the module.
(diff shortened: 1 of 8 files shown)


framework/Web/UI/TTemplateManager.php
```diff
@@ -69,17 +74,25 @@ public function getTemplateByClassName($className)
...
+	 * @return \Prado\Web\UI\TTemplate or subclass template parsed from the specified file, null if the file doesn't exist.
 	 */
-	public function getTemplateByFileName($fileName)
+	public function getTemplateByFileName($fileName, $tplClass = null, $culture = null)
 	{
-		if (($fileName = $this->getLocalizedTemplate($fileName)) !== null) {
+		if ($tplClass === null) {
+			$tplClass = $this->_defaultTemplateClass;
+		}
+		if ($tplClass != TTemplate::class && !is_subclass_of($tplClass, TTemplate::class)) {
```
ctrlaltca, line 87: We have an ITemplate interface that we could use. in place of this.
We'll probably need to add to the interface the method getIncludedFiles() used by TTemplate Manager for caching purposes

ctrlaltca (changes requested):
LGTM
What do you think about using the ITemplate interface?
This would make it easier to use templates that are not based on the prado syntax

belisoful:
So.  The issue here is that I'm running PHP 8.2 and it doesn't work with PHP-CS-Fixer.

For PHP-CS-Fixer to work correctly, it must be PHP 8.1.  I'm going to downgrade my PHP, re-fix, and re-commit the corrections.

framework/Web/UI/TTemplateManager.php
belisoful: Keying from ITemplate in the latest iteration.  ITemplate adds getIncludedFiles and the various ITemplates around PRADO have been updated with the new function returning [].

framework/Web/UI/TTemplateManager.php
```diff
@@ -69,17 +74,25 @@ public function getTemplateByClassName($className)
...
+	 * @return \Prado\Web\UI\TTemplate or subclass template parsed from the specified file, null if the file doesn't exist.
 	 */
-	public function getTemplateByFileName($fileName)
+	public function getTemplateByFileName($fileName, $tplClass = null, $culture = null)
 	{
-		if (($fileName = $this->getLocalizedTemplate($fileName)) !== null) {
+		if ($tplClass === null) {
+			$tplClass = $this->_defaultTemplateClass;
+		}
+		if ($tplClass != ITemplate::class && !is_subclass_of($tplClass, ITemplate::class)) {
```
ctrlaltca, line 87: a single instanceof check should suffice

ctrlaltca (approved):

PR #905, opened 2023-04-20
#824 updated the TComponent::instanceBehavior to init

belisoful:
Sorry about not leaving a comment.  This corrects that behaviors that are created are initialized, not just when they have a configuration.


belisoful:
This merge initializes all behaviors that are instanced after they are checked for type.

my apologies for not leaving my usual in verbose comment.

PR #906, opened 2023-04-21
#827 TTemplateManager no need to check class against an interface

belisoful:
remove remnant code from when checking against TTemplate rather than ITemplate.


belisoful:
OK.  I think I just added a branch to Prado by accident for this fix.  When it asked to branch the change, I thought it was going to be on my personal master.  live and learn.

Please delete this branch when merged.  My humble apologies for the extra branch.

Issue #851, opened 2022-12-25
TBehavior event handlers not adding at the proper priority.


belisoful:
Basically, Behaviors should implement the new IPriorityProperty merge #849 at the IBaseBehavior level.

When the behavior is attached, the priority is already set on the behavior from being added to the list of owner behaviors.  thus when the behavior is attached (and then events are added), the behavior events can be added at the behavior's priority.  this is the expected behavior.

Much of the priority stuff in TBehaviorModule can be removed as well

belisoful:
BTW, long ago Yii did away with the "enabled" flag on behaviors.  I am very glad we kept it despite the complexity I am seeing now.  The ability to enable and disable a behavior in the configuration without removing it entirely is very useful.

I see why they did it.  It simplifies things, especially some core code in checking for enabled.  For speed.

If this is something that people would like to see, removal of the Enabled flag in behaviors, let's converse.

belisoful:
I've been able to identify several strong points of the Yii implementation that should be accounted for and some weak points that are being upgraded.

@ctrlaltca The size of the change of behaviors is "big" and covers 3 active issues (and then some) but the size and scope of the upgrade to behaviors can't be helped.  it stays in the scope of the core of behaviors (I/TBaseBehavior, I/TClassBehavior, and regular I/TBehavior interfaces and implementations [for 6 total classes]).  With unit tests. 

Yii has two features Prado doesn't have: Anonymous behaviors, and Lazy Loading of behaviors.

Prado has a few differences: Behaviors have priority.  Events have priority.  Object themselves have a "behaviors" enable flag, individual behaviors have an enable flag, (stateless) Class-wide behaviors, behaviors attach on Interfaces and Traits also (Yii only does objects with parents [and config?]), dynamic events, and global events.   

New: Behavior events have behavior priority.  Behavior registered Events are optionally removed on disabling of behaviors or a behavior, by default they are removable on disable.  Yii is active all the time.  (In effect PRADO allows behaviors to be turned off in configuration without being totally removed). Multiple handlers can be placed on the same event without hacking the case of the event name. Class behaviors track all their owners by WeakReference. Behaviors are given a name property.

About the 2 Yii features not in PRADO:
 - Anonymous [numeric/indexed] Behaviors requires an update to TMap and TPriorityMap to have an index on added items with null as key (and checking the index and incrementing to not overwrite).
[...]

PR #912, opened 2023-04-27
4.2.3 General Behaviors Update... 

### LINKED ISSUE #891: TComponent:attachBehavior not checking for existing behavior name
Behaviors are overwritten.  But it should properly remove the prior behavior at the name before simply overwriting.

This works better:
```
	public function attachBehavior($name, $behavior, $priority = null)
	{
		$behavior = self::instanceBehavior($behavior);
		if ($this->_m === null) {
			$this->_m = new TPriorityMap();
		} elseif ($this->_m->contains($name)) {
			$this->_m->detachBehavior($name);
		}
		$this->_m->add($name, $behavior, $priority);
		$behavior->attach($this);
		$this->dyAttachBehavior($name, $behavior);
		return $behavior;
	}
```

BTW, Yii doesn't have priorities for their behaviors or events.  It can append or prepend only.

They do not have dynamic events to tightly bind behaviors to their owner and it doesn't have global 'fx' events.

belisoful:
This fixes #891, #851, and #838

- TClassBehavior tracks owners with TWeakList
- Attaching IBehavior to Class clones instances for each owner/instance. the standard action was to instance from a class name string or array of class/properties.
- TComponent::__destruct clears the behaviors and event handlers for better garbage collection.
- IBaseBehavior has Enabled flag.  Thus, IClassBehavior has enabled flag.  Simplifies TComponent behavior processing.
- TBaseBehavior records its unique attachment name in the owner.  IClassBehavior can only install on one name across all owners.
- Attaching a behavior over an existing name properly removes the prior behavior.
- TBehavior and TClassBehavior track the behavior enabled status of their owner by dynamic event dyEnableBehaviors and dyDisableBehaviors for turning on and off their attached event handlers.  This is eating your own dog food.
- Changing the enabled status of a component synchronizes its behaviors event handler attachments.
- Disabling an owner behaviors or disabling a single behavior will detach its registered handlers from all.
- Flag for whether to detach handlers when disabled
- EventsLog to hold a cache of the behavior events, for Closure retainment.  Now supports Closures as behavior event handlers properly.
- TBehavior tracks its owner with WeakReference.  For better garbage collection.
   - Both track the attachment status of the event handlers in their owner and owners.
- IClassBehavior file creation author properly attributed.  Class Behaviors were never a thing in that time in PRADO.
- Behaviors can implement TComponent dynamic events on owners without being self called when calling their own TComponent methods (that call the dynamic event).  Behaviors work better with its own dynamic events.
-TComponent::disableBehaviors makes special implementation for dyDisableBehaviors after Behaviors are Disabled so behaviors can act accordingly because behaviors are disabled when the behaviors' dynamic events are raised.
-TBaseBehavior implements getStrictEvents to say whether the events are mandatory to attach or optional.
- Error Messages
- Updated documentation on behaviors in TComponent and in Behaviors.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,7 +1,7 @@
 ## Version 4.2.3 - TBA
 
-BUG: Issue #824 - All system initialized behaviors call the behavior "init($config)"" similar to TModule. (belisoful)
-BUG: Issue #848 - TComponent events support Closures [anonymous functions]. (belisoful)
+ENH: Issues #824, #838, #851, #891 - General Behaviors Update: IBaseBehavior has init($config) method. TClassBehavior tracks their owners.  Behaviors attach their registered event handlers at the behavior priority.  Registered Behavior event can optionally attached and detached automatically when the behavior is enabled or disabled (default).  Behavior events() support Closures and multiple handlers per event. IBehaviors are attachable class-wide by cloning. Behaviors for behaviors has better support. (belisoful)
+ENH: ISSUE #848 - TComponent events support Closure (anonymous functions) as handlers. (belisoful)
 ENH: Issue #886 - Lists the 1st level traits of the class and its parents in TComponent::getClassHierarchy.  Class-wide behaviors support attaching to Traits as well as interfaces, classes, and their parents. (belisoful)
 ENH: Issue #845 - PHP Clone and Unserialize of TComponent objects supports behaviors. (belisoful)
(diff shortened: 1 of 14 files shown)


belisoful:
my PHP Stan get removed while swapping, the errors were because my phpStan wasn't working right.  reinstalled.  found errors.

I have spent an enormous amount of Time proofing and vetting this and the php documentation.

Unless you find something: this is ready for merge.

ctrlaltca (approved):

Issue #918, opened 2023-04-27
Spot TDbCronModuleTest test failure


ctrlaltca:
Sometimes a TDbCronModuleTest unit test fails.
```
1) TDbCronModuleTest::testLogCronTask
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'testTaskAA'
+'testTaskDD'

/home/runner/work/prado/prado/tests/unit/Util/Cron/TDbCronModuleTest.php:246
/home/runner/work/prado/prado/vendor/bin/phpunit:123
```

You can see an example here: https://github.com/pradosoft/prado/actions/runs/4822368307/jobs/8589506193

belisoful:
I've noticed this.  I have a fix for this in the up coming Cron v2 update...  I've been wildly distracted from that portion... but it's still waiting in the wings.

I think this issue come from an SQL order that isn't happening and two elements are rarely "randomly" gets switched.  The unit test could be "corrected" to look for elements in not just static places... but the new Cron code should fix this.

belisoful:
Let me clarify.  The presumption in DB Cron is that when a DB record is inserted, it will be returned in order of insertion and so the SQL is not ordered by Record UID in getting the records because of this presumption.

As we are seeing, the DB is not always returning the same order as insertion.  I don't know why.  However, the new Cron V2 does order by insertion id in this case.

belisoful:
If you are looking for a quick fix, one of the SQL SELECT statements "just" needs to be ordered by the UID.  I'm sure it would be easy to nail down given the unit test that is failing and seeing the cron method producing the wrong results.

Besides properly ordering by UID, Cron v2 lazy loads the jobs from the DB rather than loads them all.  Only those needed are loaded.  It can handle a much larger load of cron jobs.   That is to say, it makes the cron system ready for a user base rather than "just for system use" or low usage.

belisoful:
I went ahead and zinged this spot error.  I do believe this fixed the issue.

belisoful:
I'm actually 99.999% sure this was the issue.  so I'm going to close this issue.  I say that from lots of testing on cron v2.

belisoful:
ok.  didn't fix the issue.

belisoful:
ok. NOW I'm sure it's fixed.

PR #919, opened 2023-04-27
#918 TDBCron updated SQL to "Order By" to stop random test failure

belisoful:
I am taking a wild swing at this one.  I do hope this fixes the bug, but it's hard to tell when the bug is so spotty.  I think this is more of a bug with the DB than PRADO, but hey, this should fix that quirk.

I didn't think this was necessary because.... wouldn't the DB return the table in the order of insertion?

apparently, randomly, it switches rows unless explicitly ordered


belisoful:
After this is merged, we'll leave the issue open for a bit and see if the error randomly keep coming up.

Issue #925, opened 2023-04-29
TMap (and TPriorityMap) need keyOf method for getting the key with the item


belisoful:
TList has indexOf, but TMap needs an equivalent.

PR #926, opened 2023-04-29
#925 TMap::keyOf is added, additional checks on TPriorityMap

belisoful:
TPriorityMap should not just return the first instance of a value but it should depend on the priority.  higher priorities are first in the map.


belisoful:
Should these be called keyOf instead of IndexOf?

ctrlaltca:
yep, keyOf seems moreappropriate

belisoful:
Corrected.

Issue #898, opened 2023-04-18
Magic Methods need PHP Visibility conformance (aka protected setters aren't called by magic method)


belisoful:
There was recently some confusion about TList/TMap::ReadOnly being protected vs public.   I was updating the unit tests for TPriorityMap and discovered that DefaultPriority is being set as a property by magic method when it is protected.

The reason why configuration and templates can access ReadOnly when it is protected is because accessing the property is still accessible by magic methods.

Example.
```
$map = new TMap();
$map->setReadOnly(true);   // Failure due to being protected
$map->ReadOnly = true; // works because the object constructs and accesses "setReadOnly" on itself.
```

Is this something to address or not?  should it be working this way?  Or is this a feature?

Magic methods, canGetProperty, and canSetProperty would need to check the getter/setter method for public.

This is also an issue in Yii2, I think.

ctrlaltca:
It's definitely something that's breaking basic OOP principles and should be fixed.
I'm not sure why, since we should check is_callable() it from the caller context.
The real question is: how many things will break if we fix this? :smiling_face_with_tear:

belisoful:
I don't think this was an error that I introduced when i added behaviors to TComponent; it was something going on before that.

belisoful:
I'm fairly certain that Yii2 has this problem too because they do a straight method_exists check as well rather than a method_exists then method-is-public check.

I'm going to make a Prado::method_exists in the base utility object that checks if the method is public and then replace `method_exists` is TComponent with `Prado::method_exists`

belisoful:
This is a serious enough bug that I did actually report this bug to Yii as well.

I have a working solution but it doesn't fix the regression like TMap::ReadOnly and TList::ReadOnly configuration setting.  

Even when this bug is squashed, we may need to revision some elements.  Like making ReadOnly public for TList and TMap but only allowing the first set for configuration.  there would need to be a second protected "resetReadOnly" for subclasses to change ReadOnly after the first set.

this may be a big enough change to "read only" that it should only be done with community in mind.

Keep in mind that with this solution, there won't be any ReadOnly for lists until ReadOnly is made public.  

Again, maybe only the first set should be honored in public SetReadOnly?  Breaking implementations that expect it to actually do something from the subclass.

belisoful:
Processing.  one eternity later....      Basically, turning off get-set properties for protected and private properties also turns it off for the object too.

[...]

PR #929, opened 2023-04-29
#898 TList/TMap:: setReadOnly made public, only allow first set & self set

belisoful:
This resolves the issue of setting TList/TMap::ReadOnly in configuration.  while preserving access control over the property.

This converts TMap and TList Read only into a null and accepts first set and self set.

Regression.

With unit tests.

### DIFF
--- a/framework/Collections/TList.php
+++ b/framework/Collections/TList.php
@@ -99,7 +99,7 @@ class TList extends \Prado\TComponent implements \IteratorAggregate, \ArrayAcces
 	/**
 	 * This sets the read only property.
 	 */
-	protected function ensureReadOnly(): void
+	protected function collapseReadOnly(): void
 	{
 		$this->_r = (bool) $this->_r;
 	}
@@ -171,8 +171,8 @@ class TList extends \Prado\TComponent implements \IteratorAggregate, \ArrayAcces
 	 */
 	public function insertAt($index, $item)
 	{
+		$this->collapseReadOnly();
 		if (!$this->_r) {
-			$this->ensureReadOnly();
 			if ($index === $this->_c) {
 				$this->_d[$this->_c++] = $item;
 			} elseif ($index >= 0 && $index < $this->_c) {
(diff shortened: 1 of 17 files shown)


framework/Collections/TList.php
```diff
@@ -76,15 +78,30 @@ public function __construct($data = null, $readOnly = false)
...
+			throw new TInvalidOperationException('list_readonly_set', $this::class);
+		}
+	}
+
+	/**
+	 * This sets the read only property.
+	 */
+	protected function ensureReadOnly(): void
+	{
+		$this->_r = (bool) $this->_r;
```
ctrlaltca, line 104: Do i get this correctly, is this this used to force null to false if unset, or leave the actual value if already set?
The name ensureReadOnly() maybe is  a bit misleading, i would assume `$this->_r = true` given the name

ctrlaltca (approved):
Looks good to me, just a single method name looks a bit off maybe

framework/Collections/TList.php
belisoful: yes.  it's to "collapse" the ReadOnly from the null state.   I really didn't know what to call the function.  Can you suggest a better function name?  I'm at a bit of a loss.

belisoful: Sometimes, we get so lost in the function and functionality that the name becomes an oversight.  Kinda like the new "method_visible" aka new "method_exists".

belisoful: ok.  I'm thinking "collapseReadOnly".

belisoful: also, sorry about changing the name of method_accessible again.  I do think it is better as method_visible as that is what it computes....  PHP Visibility.

ctrlaltca: Quite had to name it, to be honest, since it will also "pollute" the child classes. ensureReadOnlyIsInitialized(), ensureReadOnlyHasValue() seems too complex.. a different verb like reduceReadOnly() could be used, even if reduce is usually associated to arrays.. i'm not sure.

ctrlaltca: collapseReadOnly sounds great! +1 also for method_visible

ctrlaltca (approved):

belisoful:
Corrected.

Issue #904, opened 2023-04-20
TDataReader and TDataWriter


belisoful:
We have a TTextWriter but we don't have any solid core classes for reading and writing streams of digital bit/byte-based data.  In the world of XML, HTML, and XHTML such formats may seem extraneous.  However there are many data that is written and read in bits rather than text encoded.

Reading and writing IPTC, EXIF, JPEG APP tags, and other digital formats does need such classes.  No doubt there are other applications for such utility...  including streaming data over TCP/UDP, and other digital formed data.

I have these done already and they do a few extra things like read half floats and mini floats that typical php and PHP Readers/Writers do not do.  It has push/pop for offset saving and restoring and sectioning for isolating a portion of a file as if it were its own file.   The sectioning is useful for designating a portion of the file (start offset and length) as IPTC encoded data and it will start the "file offset" at the start of the IPTC block, and act like its the end of the file at the end of its length.

Necessity is often one of the greatest sources of inspiration.

belisoful:
more utility is needed than just TDataReader and TDataWriter.

TDataStream - encapsulates a PHP Stream, standardizes stream behavior (mainly writing).
TDataReader- reads binary data in a specified endian
TDataWriter - writes binary data in a specific endian
TFileReader - initialization but with a file path.  a slightly more specific TDataReader
TFileWriter - initialization but with a file path. a slightly more specific TDataWriter
TTempWriter - specific for making a temporary file writer stream.
TMemoryWriter - specific for making memory writer stream.

TBitReader - for reading a specific number of bits at a time.
TBitWriter - for writing a specific number of bits at a time.

TDataAllocation - structured memory allocation functionality

TLZWCompressor - A common, effective, and fast compression algorithm (with expired patent)
TPackBitsCompressor - used by TIFF.

belisoful:
TTDataStream may be able to handle files is being designed for files, but I think sockets will fit and can be handled as well.  Sending binary data by socket through TDataWriter would be utilitarian.

belisoful:
This is a more interesting issue that I initially figured.  Reading and writing binary numbers in a stream could mean the steam is a resource, file, socket, I/O pipe, or process pipes each with their own slighting different open and close function.  Each are using fread and fwrite to receive and send data.

This adds a great abstraction layer to interprocess communication, which we may need if cron is going to spin off each cron job as its own process at some point.

[...]

PR #934, opened 2023-05-02
#904 TRational/TURational unit classes Prado\Util\Math

belisoful:
The Mathematical "rational" is an integer numerator divided by an integer denominator.  This is the class providing the functionality around reading, computing, and outputting the rational.

The format for the Rational is, eg,  "21/13" where 21 is the numerator and 13 is the denominator.

The unsigned rational is bound to 0..4294967295; where as the regular rational is signed and bound to [-2147483648..2147483647].

A Rational is found in EXIF and is common in its GPS "directory".  It is found there as two 32 bit integers.

The star method of this class is the Continued Fraction computation to generate the integer numerator and integer denominator from a float, to a specific tolerance.

Updated TBitHelper with specific max, min, unsigned max, and mask for 32 bit and 64 bit integers, when the values are needed and not just the system values.

### DIFF
--- a/framework/Exceptions/messages/messages.txt
+++ b/framework/Exceptions/messages/messages.txt
@@ -581,6 +581,8 @@ dbcron_property_unchangeable			= TDbCronModule.{0} is already initialized and ca
 
 timescheduler_invalid_string			= TTimeScheduler could not parse the schedule element '{0}'
 
+rational_bad_offset						= "{0}" is not a valid property accessor of TRational.
+
 bithelper_bad_fp_format					= TBitHelper cannot work with '{0}' exponent bits, '{1}' mantissa bits, '{2}' exponent bias, in a {3} bit PHP environment.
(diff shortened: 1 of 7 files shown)


belisoful:
I thought the invokable would be interesting...  I like the array access better.  I'm goin to remove the __invokable access....   wait on this merge for a moment.

belisoful:
Should the __invoke be kept or removed?  I'm on the fence.  I like that it accepts tolerance.

belisoful:
I was dreaming about it this morning: Putting a TRational as an event handler because it's invokable. 

lol.  That's the only thing about invokable that makes me pause about having TRational being invokable like that.  is it a good hack?

The invokable was implemented before access to the Rational Value as a [null] array accessor, and I like the array accessor much better than invokable (due to the implications of being invokable).

After my cup of coffee, I'm thinking of removing it.  Let me do that, and it'll be acceptable as a merge.

My thinking of invokables is that their utility is more specific than this use.

Issue #940, opened 2023-05-10
TDbCache::setValue collisions - TApplicationComponent::getClassFxEvents thrashing cache


majuca:
HI, since the version 4.2.2, we have a lot error like this:

![image](https://github.com/pradosoft/prado/assets/6562387/d3dcdc08-7725-4bce-ad24-dc414e6cad77)

An idee?

ctrlaltca:
Possibly related to issue #820 and this change in PR #841: https://github.com/pradosoft/prado/pull/841/files#diff-e89e3dcbd35b5518b965e802e9aa3d2d3ecb051a7c6fe6feaafd0928c17bc277
Is the database mysql/innodb? You should be able to get some more information on the last deadlock with ` SHOW ENGINE INNODB STATUS;`
Also, did you flush (wipe) the cache when upgrading prado version? prado-cli includes a "cache" action to perform the flush

majuca:
Indeed, I would confirm that problem is related to the issue #820. If I change the application mode to Normal instead of Performance, it seems to be ok. Yes, I'am with mariadb/innodb 10.11.2.

The cache was removed completly.

Maybe the folloing code here in the function TApplicationComponent::getClassFxEvents:

if (**$mode === TApplicationMode::Performance** || isset(Prado::$classMap[$baseClassName])) {
      $_classfx[$className] = $fx;
      $cache->set(self::APP_COMPONENT_FX_CACHE, $_classfx);
}

belisoful:
Dang.

belisoful:
It is difficult to debug that function. I have four comments:

1) Getting this function right is critical.
2) My apologies.
3) When mode is Performance the cache should always be set, so something else is going on
4) When the Mode is normal, maybe isset(Prado::$classMap[$baseClassName]) isn't finding the map classes properly.  maybe because of case sensitivity?

belisoful:
Maybe instead of one giant table with each class as a key, it should have multiple keys, one for each class?

belisoful:
@majuca The issue is that the code "shouldn't" be getting to that point.  The TApplicationSTatePersister should have had the GetClassFxEvents loaded from cache already.  That code only runs once to load the cache, and not a second time, unless the cache is manually busted.

If that code is running multiple times (as it seems to be doing in your instance), then the classFxEvents are not being cached properly.

belisoful:
[...]

PR #941, opened 2023-05-10
#940 Correcting the scope of getClassFxEvents::$_classfx

belisoful:
"Restoring order to the force." - Not Yoda

presuming scope and data that isn't there can lead to errors.


belisoful:
We may want to do a general 4.2.3 release to fix this bug in general sooner rather than later.

belisoful:
It would be nice to get the new TEventHandler into the 4.2.3 release if that is determined (due to this bug fix).   I almost have it ready for commit, it's coming within 24-72 hours.

PR #947, opened 2023-05-15
TException shouldn't be shifting off the first arg

belisoful:
When updating TException I accidentally left in an array_shift.

It shouldn't be shifting off the first argument like before.


belisoful:
When moved the $argv to the parameter list I didn't remove the shift.  This corrects the error.

Issue #948, opened 2023-05-16
TException Message vs Code, no int Code, no display of numeric error codes in error HTML


belisoful:
There is general IO to and from a Socket-stream that I have plans to support.  Sockets and Socket-streams emit integer error codes, and can emit the PHP translated message.

There is no way to set the actual integer error code of an Exception only its message.   Recent changes allowed for exception chaining.

For completeness, I'd like to set, capture and display the numeric integer error Code if there is one.  They way this could be done is by appending `"=NNN"` to the "messageCode";  where NNN is an integer.

This 100% backward compatible and only error codes in the new style of `"component_error_message=6"` will have the new functionality.  In this example the error code is "6". and the "=" will be stripped off also.

belisoful:
The presumptions about variable parameters are very much baked into PRADO.  there is nowhere to fit an actual "errorCode" except as "=NNN" part of the messageErrorCode

PR #949, opened 2023-05-16
#948 TException adds support for PHP's integer Exception::Code

belisoful:
By appending `"=NN"` to the errorMessageCode the integer error Code can be set.

This also adds the TNetworkException and TSocketException.


belisoful:
The error templates need to be updated for Error Codes but that is another revision.

ctrlaltca (commented):
This sounsd a bit too hacky.
 What if  we update the ctor like this:
```php
public function __construct($errorCode, $errorMessage, ...$args)
{
  if(!is_int($errorCode)) {
    //assume old code
    array_unshift($args, $errorMessage);
    $errorMessage = $errorCode;
    $errorCode = 0;
  }
}
```
This way the default is the new, correct case. The whole "if" block can be removed in the future

belisoful:
I actually agree with you on the hackiness.   I wasn't expecting this to be accepting as is.

Your parameter overriding is similar to the Throwable for $previous chaining of exceptions.  I love it.  I'll patch it up.

belisoful:
FYI,  I'm updating the PHPDoc for this change, so it's taking a few moments more than usual.

Also. this is the code that fits:
```
public function __construct($errorCode, $errorMessage = null, ...$args)
	{
		if(!is_int($errorCode)) {
			//assume old code
			if ($errorMessage !== null) {
				array_unshift($args, $errorMessage);
			}
			$errorMessage = $errorCode;
			$errorCode = 0;
		}
```
because there may not be a second parameter in the old style, which is fine.

belisoful:
and I'm thinking that some crazy random value that won't be a parameter instead of null.

belisoful:
OK. Ready for review.

Keep in mind that the error template might need updating with the %%ErrorCode%% token.

PR #968, opened 2023-06-05
TStreamNotificationCallback for observing fopen/file_get_contents file connection/download

belisoful:
This is an IO helper class for creating notification $context and routing  notification events from the callback 

This is depends on  #967 merge.  only merge after #967


belisoful:
I'll get that.

belisoful:
One more minor correction...

belisoful:
OK.  That's it.

PR #978, opened 2023-06-20
TComponent upgrade: raiseEvents in reverse [option], search for behavior by class

belisoful:
There are 5 upgrades to TComponent:
- __destruct ordering is corrected so behaviors are removed before unlisten. this is to properly unwind and reverses the order of __construct.
- raiseEvent can execute the handlers in reverse.  This is needed if the first handlers needs to be executed last for proper priority/importance of the result.   This adds TEventResults::EVENT_REVERSE
- asA is updated so that if a behavior name is not found it, and the $behaviorname is a class/interface, it will look for the first behavior instanceof
- isa is updated to include traits so traits can properly have global class behaviors.
-getBehaviors is updated to return all behaviors of a select class if a class is provided

this also updates and refactors the TComponent unit tests to not use behavior classes as behavior names.


belisoful:
need to update phpstan and csfixer.  Just a moment....

belisoful:
https://github.com/pradosoft/prado/issues/979

Issue #977, opened 2023-06-19
THttpRequest::onResolveRequest for application specific request => service resolution


belisoful:
In some instances, subclassing the HttpRequest isn't flexible enough to "append service resolution" functionality.

A developer may have a specific HttpRequest subclass and want to add a WebSockets layer to the Http Request...  a double sub class isn't possible.

This adds an application layer to service resolution for custom service discovery.

belisoful:
a webSocket service can be activated by designating a URL parameter, there should be a behavior (and thus an event) for manipulating the URL Parameters and providing for custom application service selection without subclassing.

A TRequestConnectionUpgrade behavior for THttpRequest for injecting the `Upgrade` http header into the $getParam.

I have the unit tests for this working.  It needs a custom parameter class for the onResolveRequest event.

there should be a way to override the serviceID resolution and cancel as well as manipulating the URL parameters.

belisoful:
I'm glad you like it.  It adds a subtlety to resolveRequest that's been needed.  The included Behavior is a great example of how it is to be used.

PR #981, opened 2023-06-21
#977 THttpRequest::OnResolveRequest and TRequestConnectionUpgrade for websockt

belisoful:
Refactor of resolveRequest


framework/Web/THttpRequest.php
```diff
@@ -767,43 +768,73 @@ protected function parseUrl()
...
+	 * @param array $serviceIDs The Service IDs of the application.
+	 * @param array $urlParams The decoded url parameters to be updated or replaced
+	 * @return array|false|string
+	 */
+	public function onResolveRequest(array $serviceIDs, array $urlParams): false|string|array
+	{
+		$results = $this->raiseEvent('onResolveRequest', $this, new THttpRequestParameter($serviceIDs, $urlParams), TEventResults::EVENT_REVERSE);
+		while (count($results)) {
+			$return = array_pop($results);
+			if (is_string($return) && in_array($return, $serviceIDs) || is_array($return) || $return === false) {
```
ctrlaltca, line 833: Please add some parenthesis to avoid confusion in operator precedence.
Where is the "null" case handled? (By returning null, the result is skipped and has no effect) I guess it just loops over all non-matched data types, including null.

belisoful: The results could be filtered for null by including the `TEventResults::EVENT_REVERSE  | TEventResults::EVENT_RESULT_FILTER` flag.  You are correct, null is simply skipped, whether filtered or just ignored.  I like adding the filter.

Also, && is always logically combined first before ||.  The operator order is specified in PHP.  with this knowledge, it's not confusing.  But if you think so, OK.

belisoful: my bad.  that's right, the event filter cannot be used because it will filter out the `false` values as well.  It is the case of simply skipping over data that doesn't match.

Only strings in the serviceIDs, arrays to update the urlParam and false.  it does run down the array of results for the first valid result

belisoful: I'm updating the parenthesis but I don't think it's needed.

ctrlaltca: > I'm updating the parenthesis but I don't think it's needed.

You are correct, it's just my paranoia due to php changing the way operators worked in the past (eg. ternary operators in 7.4)

belisoful: I know what you mean.  some things are wildly more nuanced than they appear on the surface.  ternary and order of operation can be like that.

BTW, I have this double ternary operator that is very interesting to see, at least to me when I first put it together.
eg `($endian === null) ? $endian ? 'N' : 'n' : 'l'`

Issue #982, opened 2023-06-21
Logging needs to be cleaned up


belisoful:
No one has really touched the Logger code since the start.   As part of the forking, I have a utility for grabbing the log from the child fork and bringing it back into the parent.  This requires a mergeLog function.

So I started looking at the whole Logger and comparing it to Yii.

While updating the logger, it needs profiling and a log flusher for long running processes.

belisoful:
I'm reviewing the logger....  because.     The TProcessHelper has an Application behavior for merging child process logs back into the main thread.  mergeLogs is needed but the Logging needs so much more.

Interesting things...

-TLogger needs an onFlushLogs event for the TLogRouters to patch into, rather than the TLogRouter patching into onEndRequest. ++. 
-TLogger needs to trigger onFlushLogs when onEndRequest is raised.  (this should patch into the fxTerminate global event in the coming TProcessHelper). ++
-TLogger needs a flush feature when the logs get too big. +
-TLogger needs a mergeLogs ++
-TLogger should log the PID for merging children logs (a feature of the new TProcessHelper) ++
-TLogger needs profiling (begin and end). +
-TLogRoute::Enabled property +
-a Context log message with specified global variables, and context exclusions?  Security issues here. +
-prefix for formatting log messages (filled with IP, logged in user ID, sessionID) +

PR #983, opened 2023-06-25
#982 General Logging update: Profiling, Flushing, Category/Control Filter Exclusion, Route::Enabled, TSysLogRoute, unit tests

belisoful:
- Adds Profiling with TLogger::PROFILE (for BEGIN and END) and TLogger::PROFILE_BEGIN and TLogger::PROFILE_END
- adds FlushCount to flush the log when reaching the max for long running processes.
- Categories can be Excluded in routes by designating '!' or '~' before the category name, eg. "~Prado\Web\UI" will exclude logs with that path.
-  MergeLogs to merge child logs
- TLogRoute::enabled
- Log Message Prefix
- TBrowserLogRoute colorizes the delta Times for easier visualization. Adds the TSysLogRoute
- Better PHP Doc
- Logging finally has unit tests for primary classes


belisoful:
I'd like to move the logging into its own folder for the 4.3 bump.  move the TUtfConverter into the Helpers folder too

framework/Util/TLogger.php
```diff
@@ -33,49 +43,388 @@ class TLogger extends \Prado\TComponent
...
+	/**
+	 * @param bool $flushShutdown Should onFlushLogs be a register_shutdown_function.
+	 * @since 4.2.3
+	 */
+	public function __construct(bool $flushShutdown = true)
+	{
+		if($flushShutdown) {
+			register_shutdown_function(function () {
+				$this->onFlushLogs();
+				register_shutdown_function([$this, 'onFlushLogs'], $this, true);
```
ctrlaltca, line 148: This looks strange, a register_shutdown_function() inside a register_shutdown_function()

ctrlaltca (changes requested):
Wow, this is huge!
I'm testing TFileLogroute at the moment and it seems that it doesn't get flushed to disk at the end of the request. 
My application.xml contains
```
    <module id="log" class="System.Util.TLogRouter">
      <route class="System.Util.TFileLogRoute"
        Levels="Warning, Error"
      />
      <route class="System.Util.TFirebugLogRoute"
        Levels="Info,Profile"
      />
    </module>
```
and i'm calling a single `Prado::warning('test');` from my code

framework/Util/TLogger.php
belisoful: Yii2 and other frameworks are models.   The second shutdown is to wrap up any logs entered by shutdown functions.   The second time is appended to the list of shutdown functions.

framework/Util/TFirebugLogRoute.php
```diff
@@ -87,25 +84,25 @@ protected function renderHeader()
 		return '';
 	}
 
-	protected function renderMessage($log, $info)
+	protected function renderMessage($log, $meta)
```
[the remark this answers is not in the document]
belisoful: The TFirebugLogRoute should probably turn off the prefix message with a custom formatLogMessage.  That would be easiest rather than trying to figure out the prefix fields, especially since the prefix can be customized by the application.

belisoful:
BTW, if you haven't seen the color coding of times in the TBrowserLogRoute, I suggest you check it out.

Lowest numbers are black and light in font weight.  Then blue, cyan, green, yellow, orange, and red.  The red are bolded.  This basic visualization makes it easy to identify slow parts of the code.

ctrlaltca:
> BTW, if you haven't seen the color coding of times in the TBrowserLogRoute, I suggest you check it out.
> 
> Lowest numbers are black and light in font weight. Then blue, cyan, green, yellow, orange, and red. The red are bolded. This basic visualization makes it easy to identify slow parts of the code.

Ouch
Screenshot_20230626_114026

belisoful:
float vs int! division by zero

framework/Util/TFirebugLogRoute.php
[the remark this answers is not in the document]
ctrlaltca: I agree, better remove the prefix: time ip user and session are not that useful when in a browser console

framework/Util/TLogger.php
ctrlaltca: Nice, i didn't think of that.

ctrlaltca:
> The unit test for the TFileLogRoute is recording so it may be some interaction with when things are logged...

Yep, give me some time to debug, maybe it's a mess caused by my own application's code

framework/Util/TLogger.php
belisoful: There are a lot of great ideas I pulled from other projects like this.   I also found a lot of "holes" that other (like Yii2) were missing or doing in a much more complex way than needed.     eg: capturing the pid.  I have a some solid child forking code coming up that will actually capture a forked child log, send it back to the main thread and log it there.  So knowing the pid of the log will become important.

belisoful:
I fixed the copied comment, divide by zero bug, and the TFirebugLogRoute printing the prefix when it shouldn't.

belisoful:
TLogRouter is acting weird.  When I put TFileLogRoute second it gets added but then doesn't show up during processing.  When first, it just doesn't get called.

belisoful:
TLogger has a list of unflushed logs, and TLogRoute retains flushed logs until final or count is high enough.

belisoful:
I see what is going on.  There are not logs and so the final flush is not being sent.

````
	public function onFlushLogs(mixed $sender = null, mixed $final = null)
	{
		if ($this->_flushing || !count($this->_logs)) {
			return;
		}
```

ctrlaltca:
> TLogger has a list of unflushed logs, and TLogRoute retains flushed logs until final or count is high enough.

Yep, but TLogger::onFlushLogs() will return early if `!count($this->_logs)`, even if the route still has logs to flush

belisoful:
I'm fixing this now....

belisoful:
The TBrowserLogRoute was working for me without the float 0.0 check (but as an int).  it should work for you now too.

The Category and Control filter function has some awesome exclusion code.  You can include a specific namespace and then exclude another.  by adding '!' or '~' as a prefix to the category/control it will be excluded rather than included.   Interestingly, if nothing is explicitly included all are included so the exclusion can be used even when nothing has been explicitly included.

ctrlaltca (approved):
Everything looks good now, thank you!

belisoful:
All this because I needed a "mergeLog" function for children logs.

Issue #984, opened 2023-06-26
TEventSubscription for attaching and detaching event handlers


belisoful:
I'd like to have a class that when given an object, its event, and a handler (with optional priority), it will remove the event when the class destructs.

Short term event handler attachments would be useful.

belisoful:
omg.  I found a third PHPStan bug:  https://github.com/phpstan/phpstan/issues/9519

belisoful:
BTW, This is ending up systematizing the newer collection classes.  So IWeakCollection, IPriorityCollection, ICollectionFilter (for the IO filtering on the Weak collections) for each respective style of collection.  

I'm also adding a TMap::add return value so that the key is returned.  This way if the array key is null (for appending via []), then the new unknown (new) key is returned.  This changes the TPriorityMap::add which returns the priority.  I think returning the key is more important.

PR #985, opened 2023-07-04
#984 TEventSubscription for temporary handler in an event

belisoful:
This adds a second order functionality of temporary items in lists/arrays.  When the subscription is dereferenced, the item is removed from the list.   I wanted this for temporary event handlers for the "interrupt" and "terminate" linux process signals.  This makes use of the Collection classes in a most utilitarian way and found a few places to upgrade.

1- TArraySubscription uses pass by reference and can do subscriptions with arrays and ArrayAccess objects.
2- TCollectionSubscription uses normal object references for ArrayAccess to prevent original variable manipulation
3- TEventSubscription use a TComponent and an event name.  

- Adds TArraySubscription (arrays & ArrayAccess, pass by reference), TCollectionSubscription (ArrayAccess, normal arguments w/o pass by reference), and TEventSubscription (for events in a TComponent).
- Adds IPriorityCollection, IWeakCollection, and ICollectionFilter (for manipulating items for storage, like "weak" classes can make the objects WeakReference).
- TMap::add returns the key for when the key is null and key is set on append.
- TPriorityList::insertAt returns priority of the item.
- Regression
- unit tests for TArraySubscription, TCollectionSubscription, and TEventSubscription.
- TArrayHelper for determining if an array is a list.


belisoful:
@ctrlaltca Do you have any idea what this unrelated issue with the XLIFF DOMNode `wholeText` could be?  

I'm not having any problems on my local machine.

belisoful:
dunno why the first run of the updated PHPStan wasn't giving the error.  I'm getting the error now with the latest PHPStan now.

ctrlaltca:
Master contains a fix, please rebase

Issue #973, opened 2023-06-18
Serving a Prado application with the built in PHP Web Server for dev purposes


belisoful:
Other platforms encapsulate the PHP built in dev web server.  Needing to install a web server with PRADO is a very steep unstated requirement.

`php -S -t /path/to/app` will start the built-in web server for PHP.  This is a little known feature of PHP we can forward to our developers.

PR #988, opened 2023-08-17
#973 PHP Web Server Action for built-in app development

belisoful:
- Adds the PHP Dev-Test Web Server Action to serve the application.  This has the feature of abiding by the "Quiet" flags at different levels.  I haven't seen any other built-in PHP Web Server [Yii, Laravel] do the quiet thing. Minor feature.
- Removed "-" from CLI Text action, it was confusing
- TStdOutWriter.  This fopens  'php://stdout' when processing a web page b/c STDOUT is not available.
- Corrected TOutputWriter to write to the output rather than stdout.
- Added README.md text for starting the PHP Web Server Action after installing the application.


framework/IO/TOutputWriter.php
```diff
@@ -12,23 +12,28 @@
...
 	/**
 	 * Flushes the content that has been written.
 	 * @return string the content being flushed
 	 */
 	public function flush()
 	{
 		$str = parent::flush();
-		fwrite(STDOUT, $str);
-		flush();
+		echo $str;
```
ctrlaltca, line 36: Is this intentional? the docblock still refers to "fwrite".
Also, afaik `STDOUT` bypassed the output buffer (`ob_start()`), while `echo` doesn't

framework/Shell/Actions/TWebServerAction.php
```diff
@@ -0,0 +1,374 @@
...
+
+	/**
+	 * This option is only used when no network interface is specified.
+	 * @param mixed $value
+	 * @return bool Respond on all network addresses.
+	 * @return static The current object.
+	 */
+	public function setAll($value): static
+	{
+		if (!$value) {
```
ctrlaltca, line 110: I guess you want to check is `$value === null`, otherway a call to `setAll(false)` will set `$this->_all` to true.

framework/Shell/Actions/TWebServerAction.php
```diff
@@ -0,0 +1,374 @@
...
+	public function getPort(): int
+	{
+		return $this->_port;
+	}
+
+	/**
+	 * @param null|int|string $address The port to serve pages, default 8080.
+	 * @return static The current object.
+	 */
+	public function setPort($address): static
```
ctrlaltca, line 190: please rename the parameter to $port

ctrlaltca (commented):
Hi, as always awesome addition!
Can you please check the few comments i left?

framework/Shell/Actions/TWebServerAction.php
belisoful: Good point.  I'll correct this.

framework/Shell/Actions/TWebServerAction.php
belisoful: indeed.

framework/IO/TOutputWriter.php
belisoful: The original TOutputWriter was incorrectly writing to STDOUT rather than "Output".  The new TStdOutWriter is the direct to STDOUT class.  This was originally my error in TOutputWriter, my apologies.  I didn't know the precise details between STDOUT and "Output".

I've been studying PHP IO for a new Prado IO package and this is the correction.

For CLI, STDOUT and "Output" is the same (hence the confusion) but for PHP web pages, STDOUT is not defined and everything goes through "Output".   For web pages, this means that 'php://stdout' must be opened to have STDOUT.   Usually webpages don't need nor use STDOUT.  But for the Built-in PHP web server, we allow pages to open STDOUT to print to the CLI (eg, TStdOutLogRoute).

I'll make sure the doc is correct.

belisoful:
I sat on this to "bake" for a few weeks to make sure everything needed was included.  The quiet options were added because of this integration time.

You have no idea how glad I am that you are here to check and validate the code.  I appreciate these code reviews immensely.

belisoful:
my hope is that this PR will help people get started with PRADO faster.

framework/IO/TOutputWriter.php
belisoful: The documentation is updated for clarity.

belisoful:
I think that's it unless you find more.

ctrlaltca (approved):

Issue #976, opened 2023-06-19
TComponent needs __callStatic to be behavior enabled 


belisoful:
TComponent::__callStatic could search the _um list of global behaviors for the static method (and each parent likewise).

belisoful:
Dang.  I found another PHPStan bug that impacts this enhancement.

https://github.com/phpstan/phpstan/issues/9465

belisoful:
The way behaviors are implemented in PRADO allows for static calls to pre-attached behaviors that other platforms can't do.  This is quite unique and should implemented for maximum potential.  The new ISingleton also enables static calls to behaviors attached to a singleton as well.

PR #989, opened 2023-08-23
#976 Adds TComponent::__callStatic, Update the ReadMe

belisoful:
- This forwards unfound static methods to a class's class-wide behaviors.  If an object is an ISingleton, the singleton's enabled behaviors are also checked.
- Added TUnknownMethodException, __call and __callStatic use TUnknownMethodException rather than TApplicationException.
- TApplication implements ISingleton so it can handle static methods with the singleton behaviors rather than just the class-wide behaviors.
- refactored unit tests -Updated the Readme with new features and how to add composer extensions to an application.  (i know there aren't many available yet, but this is how we get community support with composer extensions)


belisoful:
FYI, This is a unique feature of PRADO because Prado has Class-wide Behaviors that are installed on classes [and interfaces, and traits].  That list of "default" behaviors for new instances acts as the reserve for magic method static class function calls in default behaviors.

I recently added the ISingleton interface.  The __callStatic method can also look at the singleton's attached behaviors as well.  There are other classes that could implement the ISingleton.

belisoful:
I think this is ready.

belisoful:
Regarding the ReadMe, It lists a few new interesting features (like behaviors, cron, and RBAC permissions) and adds a section on PRADO Composer Extensions.  It also removes that other frameworks don't have features because there are so many frameworks that do so much these days. 

I think we could be doing more for the community in promoting composer extensions and bringing more visibility to that feature and the community.

belisoful:
Are there any questions I can answer about this PR?

ctrlaltca (approved):

ctrlaltca:
> Are there any questions I can answer about this PR?

Sorry, i saw you were adding commits and i was not sure you were done on it.

belisoful:
That's ok.  I think you actually have a really great gut feeling about these things.  usually when there is delay, even without knowing it, there is reason.  it may need a little minor tweak here or there.  Don't fret.  We got this.

belisoful:
Also, if you are interested in providing modules from your source and archive to make into extensions, I would be willing to help move some it forward.  Having a catalog of usable extensions would be great to have.  There are several extensions I'd like to throw out there, including a Wordpress Integration plugin.  

If a PRADO Wordpress Integration extension is something you'd be interested in helping with, let me know and I'll make it a priority to get it up on GitHub.

Issue #1004, opened 2025-09-02
Add support for Content Security Policy header


ctrlaltca:
Content Security Policy (CSP) is a security header better documented at https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP
In order to use it with a Prado app it's currently necessary to specify `unsafe-inline` to let the browser load script and styles that Prado places inline inside the page.
Three possible alternatives:
 1. move every script/style block to an external file
 2. calculate an hash for each script/style block and add it to the header
 3. create a per-response nonce value and add it to each script/style block.

The first alternative is probably the best generic solution, but in order to implement this we would need to create a few new javascript files for each request (eg. for beginScripts and endScripts registered by `TClientScripManager`), handle their lifecycle (aka delete them after load) and ensure they can't be reached by other clients (they may contain private data).

The second alternative requires Prado to calculate the `base64(sha256($content)) ` for every script and style block, making all of them available to be added to the CSP header. Additional `integrity` attributes (content hashes) must be present on external scripts. This is solution meant to be used for static scripts, eg. while importing a common library from a CDN.

The third alternative requires Prado to generate a nonce (that is a random string) to be added to the CSP header and then add a "nonce" attribute on every script/style block. 

We should possibly move whatever is possible to external scripts (solution 1), create some hooks to permit the generation of hashes for script/style blocks (solution 2) and add support for optional nonces wherever we output a script/style blocks (solution 3).

Once this is in place, a new module can be created taking care of:
 * collect the CSP configuration
 * output the CSP header
 * implement solution 2 (hashes) 
 * implement solution 3 (nonces)

PR #1005, opened 2025-09-02
[WIP] Branch to implement #1004

ctrlaltca:
This branch will contain work to implement support for Content Security Policy (#1004)
The PR will ensure code is tested while working on it

By now the following steps has been committed:
 * move all `<script>` tag rendering in a single place


belisoful:
What do you think about having a CSP Reporter Service? for receiving CSP violations from browsers?

The THttpHeadersManager could have a bool|string flag with true/false/Auto that installs the TCSPReporterService.  When the "report-to" (and "report-uri") policy/directive is present, "Auto" adds the service.  true always installs the CSP Reporter Service regardless of using "report-to".  It may help to debug CSP violations as they occur.

The TCSPReporterService should log the report and trigger its "on" event for handling the error.  It should have a static method for retrieving the application TCSPReporterService if it's available; so modules-etc can easily install an event handler on a CSP Report.

Also, part of page lockdown process of CSP is "integrity" for external JS and style.  I've already put a shell script into the prado-dev-tools for looking at local and remote files and getting their SHA hash for "integrity".  The issue is that stylesheets would also need a wrapper like TJavascriptAsset and to centralize the stylesheet render.

ganiuszka:
I think this is a really good idea. I especially like the part about the "on" event, it makes it very flexible :+1:

belisoful:
The expanded version of this requires a few positive changes to the framework.
 - TEventParameter should be able to lock its Parameter with a ReadOnly.  Obviously, event handlers shouldn't be able to edit the violation.
 - The TCspReporterService being programmatically added requires an onConfigurationComplete after config before service resolution.  It allows for other modules to interact with THttpHeadersManager.

This brings up a really good point: THttpResponse is the primary "source of truth" regarding headers. THttpHeaderManager is a secondary process.  It makes sense to establish THttpHeaderManager as the main centralized point for headers, and maybe cookies. This may be worth investigating after this PR.

belisoful:
@ctrlaltca When you have a moment, `csp` needs to be rebased to Head to include the new TApplication Service API.  The initial addition of the services API was not quite enough.  The Service API needed to be fully fleshed out.

The updated branch can be filled out with the TApplication Service API regarding the TCspReporterService.

ctrlaltca:
> @ctrlaltca When you have a moment, `csp` needs to be rebased to Head to include the new TApplication Service API. 

Done

belisoful:
If you rebase `csp` to the current `master`, we'll be able to add the functional tests in playwright, which is needed before merging.

ctrlaltca:
Rebased to master. Is THttpHeaderName supposed to be the only head-related file outside Web\HttpHeaders?

belisoful:
Yes.  THttpHeaderName is more universal than just isolated to the http headers (folder), and it is similar to TMediaType and TContentDisposition.  THttpResponse uses the header names without context to the header manager.  I think of the folder HttpHeaders to be more associated with the manager than explicitly for everything related to HTTP Headers.  It is the Manager and the header classes that took up "file/visual" space in Web.

When this gets merged into master, my plan is to implement the use of THttpHeaderName in the framework.  csp branch uses it in isolation to limit the scope.  I'd like to use TMediaType in the framework too, after merge.  Again,  the wide usage is not in csp because the scope would be much bigger.

We can move it into HttpHeaders folder, but it has the same level of priority as THttpRequest in terms of Web.

PR #1031, opened 2026-04-01
Corrects TException::addMessageFile Regex

### LINKED ISSUE #1029: TException::addMessageFile regex errors
This regex is incorrect:
https://github.com/pradosoft/prado/blame/357b8c9dc902714a555080b7fa414bfe39839e2d/framework/Exceptions/TException.php#L104

This $matches[2] should be $matches[3]:
https://github.com/pradosoft/prado/blob/357b8c9dc902714a555080b7fa414bfe39839e2d/framework/Exceptions/TException.php#L106

It's not been an obvious bug because PRADO Plugins are really used much (or at all?).

belisoful:
PHPStan didn't catch this malformed regex, probably because it kinda worked.

Fixes #1029.

### DIFF
--- a/framework/Exceptions/TException.php
+++ b/framework/Exceptions/TException.php
@@ -101,9 +101,9 @@ class TException extends \Exception
 	 */
 	public static function addMessageFile($file)
 	{
-		if (preg_match('/^(.*)(-.{2.4})?\.(.{2,4})$/', $file, $matching)) {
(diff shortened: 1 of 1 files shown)


ctrlaltca (approved):
Nice catch

belisoful:
This corrected regex is very important for error messages of plugins.

ctrlaltca:
Yep, i saw the old code didn't pick up the file extension correctly.

PR #1032, opened 2026-04-02
IUserManager/TUserManager update

belisoful:
Adds IUserManager::onFinalizeUser.

Update TUserManager: 
- xml config -> php config (xml converted to php)
- removed xml config (only one config method to support)
- getUniqueRoles
- supports all hash algorithms of PHP `hash`, when available
- much better unit tests

The purpose is to start moving in the direction of PHP style configurations in total.
This converts XML into PHP and then the configuration is processed in PHP.  We only need one configuration method.
TUserManager should support all the hash functions found in `hash`; beyond the basic PRADO provided modes.
This also expands the unit tests.

### DIFF
--- a/framework/Security/IUserManager.php
+++ b/framework/Security/IUserManager.php
@@ -26,12 +26,14 @@ interface IUserManager
 	 * @return string name for a guest user.
 	 */
 	public function getGuestName();
+
 	/**
 	 * Returns a user instance given the user name.
 	 * @param null|string $username user name, null if it is a guest.
-	 * @return TUser the user instance, null if the specified username is not in the user database.
+	 * @return ?TUser the user instance, null if the specified username is not in the user database.
 	 */
 	public function getUser($username = null);
+
 	/**
 	 * Returns a user instance according to auth data stored in a cookie.
 	 * @param \Prado\Web\THttpCookie $cookie the cookie storing user authentication information
@@ -39,12 +41,14 @@ interface IUserManager
 	 * @since 3.1.1
 	 */
 	public function getUserFromCookie($cookie);
+
 	/**
 	 * Saves user auth data into a cookie.
 	 * @param \Prado\Web\THttpCookie $cookie the cookie to receive the user auth data.
 	 * @since 3.1.1
 	 */
 	public function saveUserToCookie($cookie);
+
 	/**
 	 * Validates if the username and password are correct.
 	 * @param string $username user name
@@ -52,4 +56,12 @@ interface IUserManager
 	 * @return bool true if validation is successful, false otherwise.
 	 */
 	public function validateUser($username, #[\SensitiveParameter] $password);
+
+	/**
+	 * Provides for finalizing a user object after it is created but before its returned
+	 * from the implementing User Manager.
+	 * @param TUser $user The user to finalize.
+	 * @since 4.3.3
+	 */
+	public function onFinalizeUser($user);
 }
--- a/framework/Security/IUserManager.php
+++ b/framework/Security/IUserManager.php
@@ -48,6 +48,14 @@ interface IUserManager
 	public function getUserFromCookie($cookie);
 
 	/**
+	 * Provides for finalizing a user object after it is created but before its returned
+	 * from the implementing User Manager.
+	 * @param TUser $user The user to finalize.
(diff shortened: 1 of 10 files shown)


belisoful:
I'm still considering adding getUniqueRoles to the IUserManager.  It is application data that other modules may need and is built into TUser.

belisoful:
This is ready for review.

Thoughts on integrating and expanding the IUserManager and PRADO implementations?

TUserManager and TDbUserManager haven't had a thorough review in a while.  It needs it given the Composer Extensions feature may have Extensions that need this vital information about the system.

framework/Security/IUserManager.php
```diff
@@ -22,34 +22,66 @@
...
+	/**
+	 * Returns the unique roles in the application. If there are none, `return [];`
+	 * @return array The unique roles in the User Manager.
+	 * @since 4.3.3
+	 */
+	public function getUniqueRoles();
+
+	/**
+	 * Returns the number of unique roles in the application. This may be handy if there
+	 * are a lot of rolls.
```
ctrlaltca: are a lot of rol**e**s

framework/Security/TUserManager.php
```diff
@@ -77,7 +84,11 @@ class TUserManager extends \Prado\TModule implements IUserManager
 	 */
 	private $_users = [];
 	/**
-	 * @var array list of roles managed by this module
+	 * @var array list of unique roles managed by this module
+	 */
+	private $_uniqueRoles = [];
+	/**
+	 * @var array list of users and their rolls $_roles[user][(rolls)]
```
ctrlaltca: and their rol**e**s $_roles[user][(rol**e**s)]

ctrlaltca (commented):
Minor typos.
All looks good to me, but i don't get the idea behind uniqueRoles.
Do you mean to add some validation so that get/setRoles() only accepts valid roles defined in uniqueRoles?
Also, I'm a bit scared of adding new abstract methods in a minor release, eg. all my apps would break since they implement a custom UserManager

framework/Security/IUserManager.php
belisoful: Sometimes a bit of "word invisibility" comes through.  🙏

belisoful:
I had concern about the IUserManager changes too (vs a point release).  I wanted to put this forward, certainly as a discussion.  I'll roll this back a bit for the Point release, and put together an update for the next minor point release.

_The point of Unique Roles is for different parts of the system to know and understand RBAC within the application_.

The reason why UniqueRoles is important is because of Role Based Access Control (RBAC). The PRADO Permission system runs on roles.  Administrators should have the users and their roles in a UI, at least at some point, to control their access and roles.  **It's kind of hard to add a role to a user if the page/app doesn't know what roles there are.**

Typically, I'd put the roles in an Application Db Parameter, but this also entirely disconnects role information from the User Manager and the User Objects.

During development in the past, i was consistently running into "Which Roles are in the system?"

I am considering a trait to make up the difference in IUserManager for entirely custom User Manager Classes.  This way, use the trait and the upgrade errors are easily fixed.

Let me noodle on this a bit.  I know this is ambitious, we need to expand PRADO in ways to support extensions better.

framework/Security/TUserManager.php
belisoful: I'm trying to document the property format.  I'll make the correction.

framework/Security/IUserManager.php
belisoful: fixed

framework/Security/TUserManager.php
belisoful: corrected.

belisoful:
I was in deep thought about the role that Roles play for Users and Permissions.  There is a new idea...   give me a moment to work on it.  It'll put the User Managers back in their place.   I was trying to fit Roles into the User Managers, and while they could go there, they deserve their own Manager.

belisoful:
I have a new TRoleManager and its implementations: TConfigRoleManager -> TParameterRoleManager, and TDbRoleManager.  It's about 30% of the way there.  it still needs unit tests, maybe a 4th (basic) implementation to automatically inject the TUserManager defined roles (eg TUserRoleManager?).  unit tests, and proofing.

PR #1033, opened 2026-04-03
XML clean up

belisoful:
Moved a few functions.
used `self::` rather than `TXmlElement::`.
adds different types of overloads to ArrayAccess.
A few more DOM Node properties and methods.
Using Function calls rather than the object properties, for better long term maintenance.
More Unit Tests.

### DIFF
--- a/framework/Xml/TXmlElement.php
+++ b/framework/Xml/TXmlElement.php
@@ -394,6 +394,26 @@ class TXmlElement extends \Prado\TComponent implements \IteratorAggregate, \Arra
 		return $results;
 	}
 
+	/*
+	 * This selects the children elements with a tag name from $xmlNode.
+	 * Each selected elements' attributes (as an array) are saved within the
+	 * returned array.  Module Configuration for XML => PHP uses this for arrays
+	 * of nodes.
+	 * @param TXmlElement $xmlNode The XML Element with children to search.
+	 * @param string	  $childTagName The tagName for selecting children by tag.
+	 * @return array An array of selected child nodes' attributes array.
+	 * @since 4.3.3
+	 */
+	public function getElementsAttrArrayByTagName(string $tagName): array
+	{
+		$results = [];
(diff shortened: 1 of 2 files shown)


framework/Xml/TXmlElement.php
```diff
@@ -167,7 +189,7 @@ public function getValue(): ?string
 	 * Sets the text content enclosed between opening and closing tags of this element.
 	 * @param ?string $value Text content of this element
 	 */
-	public function setValue(?string $value): void
```
ctrlaltca, line 170: please fix the type hint in the phpdoc, too

ctrlaltca (approved):
Looks good to me, just a minor phpdoc fix if you want

framework/Xml/TXmlElement.php
belisoful: I removed the ?string type def because of `if ($value !== null) {$this->_value = TPropertyValue::ensureString($value);`.

my thought was to not throw (PHP Function Parameter Type) errors if its something other than a string, and let TPropertyValue::ensureString do the heavy lifting.  This is more graceful, as it was before the recent XML update.

It's still a `?string`, and the phpdoc says so.

So my question, should we upgrade the function parameter type and remove the TPropertyValue::ensureString as the [current] updated code? or keep the original, without the function parameter type but with the TPropertyValue::ensureString, as original?

belisoful:
The first pass was basically functional.  this second pass, on Xml, is just to touch up the things i originally missed.

belisoful:
There are some updates to #1032 that depend on this PR.

framework/Xml/TXmlElement.php
ctrlaltca: I'd keep the origin to avoid breaking existing code relying on `TPropertyValue::ensureString()` sanitization.
The $value parameter can remain `mixed`, and i guess the phpdoc can reflect this, too.
Not a big one anyway, I'd let you decide what's best

PR #1036, opened 2026-04-05
Adds Firebird support to ActiveRecord, DB2 & Firebird CmdBuilder

belisoful:
While testing out what these Coding Agents could do, I asked it to "fill out" the Data directory to support the missing databases.  It spent about 15 minutes and came up with this. The next prompt was to generate the unit tests and for the unimplemented unit tests (mssql and sqlite) to be filled in.

The initial commit was the exact output of Claude.
The second commit had minor corrections with php-cs-fixer, and one method was hallucinated.
The third commit was the exact output for the unit tests.

Adds:
 - Firebird scaffolding for Active Record
 - Firebird Command Builder
 - IBM DB2 Command Builder
 - CLI ActiveRecord Shell Action updated with each supported database
 - Command Builder Tests for: Firebird, IBM DB2, Oracle (oci)
 - Column Tests for: Firebird, IBM DB2, Oracle (oci)

Implements:
 - The skipped MSSQL Column unit tests (likely skipped because of lack of environment to test)
 - The incomplete Sqlite Column unit tests

And the total cost: ~$0.75 in tokens.   A bottle of soda costs more than this PR.  Indeed,  I spend more time preparing the text of this PR than it took to generate the code.

What do you think? How does it look?  How do we test this?


ctrlaltca:
It's a nice exercise, but I'm not sure how many prado users are actually running firebase/db2, so probably of reduced usefulness.
Still, from a quick overview the code looks correct and it's probably working, too.
Tests are nice but useless, since they are not plugged in the workflow.
Can you try ask it to pipe the tests into the github actions workflow at https://github.com/pradosoft/prado/blob/master/.github/workflows/prado.yml?
If it works and tests are successful..

belisoful:
I figured that "reduced usefulness" was why it was never completed.  It was always a little thing that irked me for being incomplete and i wanted to see what it _could_ do.

@ctrlaltca the $20/mo plan on Claude is a lot more than the Free plan, but wow. it can run through tokens fast.  I see why people opt for the $200/mo plan.  The hard limit resets on April 8th, and i'll asking it to incorporate the database tests into the yml.  100%.

maybe i should have put the tokens towards another aspect of PRADO, but i wanted something obscure and less important to see what it could do first.

I'm going to let this sit as a Draft PR until we see what it can do to incorporate these db tests into yml, which would be nice!  It may take a few days.  This won't be a perma-PR.

belisoful:
Gap Analysis of Database support:
> Practical recommendation:
> - MariaDB is the most impactful gap. It's arguably more common than MySQL in modern PHP hosting environments, and the behavioral divergence (especially around AUTO_INCREMENT vs SEQUENCE, and RETURNING) is meaningful enough that a TMariaDbMetaData subclass of TMysqlMetaData makes sense rather than assuming MySQL compatibility.
> - Snowflake would be the second addition if the target use case extends to data warehousing.
>Everything else  is either already covered by an existing driver or too niche (SAP HANA, Informix, CockroachDB, DuckDB) to justify the implementation cost.

belisoful:
@ctrlaltca Firebird, IBM DB2, Oracle, and SQL Server are verified working.

ctrlaltca:
Only DB2 tests seems to actually work.
All Oracle tests are marked as "skipped".
The firebird test is not even able to create the database: `I/O error during "open" operation for file "/firebird/data/prado_unitest.fdb"`

belisoful:
I'm surprised it passed.  It's getting there.

belisoful:
This is it- _We are living in the future._  This PR actually works and everything is tested.  All unit tests are performed and passing. This is a branch to archive.  **I believe this PR is ready.**

About 99.95% of this was done by Claude Code, with error output feedback and corrections.  The last 0.05% - It could have figured out that its unit test wasn't setting the db connection active to true; the unit test itself wasn't correct.  However, I didn't want to spend tokens on something so simple.

I am really pleased with MiniMax 2.7 @ (opencode.ai)[opencode.ai] as well, at this time.

ctrlaltca (approved):
Looks good to me!

PR #1037, opened 2026-04-08
Exceptions Unit Tests and Message comments

belisoful:
Minor touch up on Exceptions.
 - Skip message file comments; both `#` and `;` are valid comments.
 - Fixed a minor error in THttpException.
 - Adds a script for synchronizing the different languages messages (no translation).
 - Adds basic Unit tests for Exceptions.


bin/sync_messages.php
```diff
@@ -0,0 +1,133 @@
+#!/usr/bin/env php
```
ctrlaltca: I'm not sure we want to put an executable script inside the framework/ folder.

framework/Exceptions/TException.php
```diff
@@ -80,7 +82,7 @@ public function __construct($errorCode, $errorMessage = null, ...$args)
 			$errorCode = 0;
 		}
 		$this->_errorCode = $errorMessage;
-		$errorMessage = $this->translateErrorMessage($errorMessage);
+		$message = $this->translateErrorMessage($errorMessage);
```
ctrlaltca: Why create a new variable and keep the old variable intact if it's unused?

framework/Exceptions/THttpException.php
```diff
@@ -47,7 +46,8 @@ public function __construct($statusCode, $errorMessage)
 		for ($i = 0; $i < $n; ++$i) {
 			$tokens['{' . $i . '}'] = TPropertyValue::ensureString($args[$i]);
 		}
-		parent::__construct(strtr($errorMessage, $tokens));
+		parent::__construct(strtr($message, $tokens));
+		$this->setErrorCode($errorMessage);
```
ctrlaltca: I don't get what changes by moving `setErrorCode()` at the end after the parent ctor. The remaining logic seems to stay the same

ctrlaltca (changes requested):

bin/sync_messages.php
belisoful: What do you recommend?  Where could it live? maybe under "tests/"? bin? This is a script we've been needing for synchronization of error messages for a while.  I just didn't know here to put it.  or if it should just be a txt without execution.

framework/Exceptions/THttpException.php
belisoful: If that is not moved, the errorCode becomes the errorMessage, incorrectly.  It's been failing the unit test, but we didn't know because there was no unit test.  If you move the setErrorCode to above, it'll fail.

It's because the THttpException translates the messages itself rather then allowing its parent to do the translating.

framework/Exceptions/TException.php
belisoful: My thought was so people wouldn't immediately "jump" to the conclusion that it was the same data as what was passed as the parameter.  It's to tame the spaghetti.  In essence, its to show that the parameter $errorMessage is no longer used "as is."  It's already doing quite a lot with parameter manipulation.  It's one less variable for people to mentally keep track of "mutating" within the function.

Generally, it's bad practice to use the same variable name for different data. It could be called `$translatedMessage`. 

But really, it's your choice.  To me, it's complex enough that something like this does reduce the mental load to process this little code.

framework/Exceptions/THttpException.php
belisoful: Refactored.  Not a problem.

framework/Exceptions/TException.php
belisoful: The translateErrorMessage was moved, and variable name reverted. I hope this provides for marginally better readability.

belisoful: I'll put this back.  It's fine.

bin/sync_messages.php
ctrlaltca: What about making it `extends TShellAction` so it integrates with prado-cli?
Otherway, just take the quicker route and move it to `bin/`
I just want to avoid possible problems for people (wrongly) copying framework/ on a public-accessible web directory.

belisoful: Here is an idea...  can we make a `prado-dev-tools` repo that is marked "--dev" in prado?  This would be a perfect place for the sync_messages.php and other such custom dev tools.

The though crossed my mind for a "bin-dev" folder, but `prado-dev-tools` is much cleaner.

Otherwise, let's drop this into bin for the moment.

ctrlaltca: You should be a full-org admin now, please check if you can create the repo

belisoful: Exercising _caution_...  With authority comes great responsibility and care.

BTW, the prado-composer-extension PR is basically waiting on the Agents update so it has a proper version of AGENTS.md.

belisoful: My account is a member of both pradosoft groups, but it still says there aren't enough permissions in the "New Repo" screen.

ctrlaltca: Can you check again now please?

belisoful: It's enabled.  🙏

belisoful: This script has been moved to `prado-dev-tools`.  Which is currently private in the PRADO group until it can be reviewed. When approved, we'll make it public, do a first release. then it'll be added to PRADO as a --dev require.

 I'll remove this script from this PR shortly, and hopefully it'll good enough to merge.

We may want to make Prado-dev-tools into a full PRADO 4 Extension that provides some UI developer overlays, with proper roles/permissions.

ctrlaltca: Nice :+1:

ctrlaltca (approved):

PR #1055, opened 2026-04-13
Completing the Web.UI Unit Tests

belisoful:
This is an effort to reduce the number of unit tests that are incomplete.

It required a few tweaks to abstract PHP functions so subclasses could override them for testing.

This is a Squash and Merge PR.


ctrlaltca (approved):
Nice work!

PR #1057, opened 2026-04-14
Refactor header handling in TErrorHandler

belisoful:


framework/Exceptions/TErrorHandler.php
```diff
@@ -137,7 +137,7 @@ public function handleError($sender, $param)
 				$response->clear();
 			}
 			if (!headers_sent()) {
-				header('Content-Type: text/html; charset=UTF-8');
+				$this->getResponse()->appendHeader('Content-Type: text/html; charset=UTF-8');
```
ctrlaltca: A response object may not exists, that's why header() is called directly.
See the $response variable a few lines before

framework/Exceptions/TErrorHandler.php
```diff
@@ -240,7 +240,7 @@ protected function handleRecursiveError($exception)
 			echo "</body></html>";
 		} else {
 			error_log("Error happened while processing an existing error:\n" . $exception->__toString());
-			header('HTTP/1.0 500 Internal Error');
+			$this->getResponse()->appendHeader('HTTP/1.0 500 Internal Error');
```
ctrlaltca: A response object may not exists, that's why header() is called directly

belisoful: great call! oh, assumptions....    I'll fix that.

framework/Web/UI/ActiveControls/TCallbackErrorHandler.php
```diff
@@ -61,7 +61,7 @@ protected function displayException($exception)
 			$content->write($trace);
 		} else {
 			error_log("Error happened while processing an existing error:\n" . $exception->__toString());
-			header('HTTP/1.0 500 Internal Server Error', true, 500);
```
belisoful, line 64: I'm going to assume that this may have the same issue.  specifically in a unit test environment.

framework/Exceptions/TErrorHandler.php
belisoful: this indicates that maybe appendHeader might be better as static, but after quick review that's not viable for subclassing via a set static class.  THttpResponse::appendHeader would call the class not a sub-class from within the error handlers.

Done

ctrlaltca (approved):

PR #1059, opened 2026-04-14
ActiveControl/TClientScriptManager update

belisoful:
- Adds removeAttribute to TCallbackClientScript.php and JS Prado.Element.
- TActiveLinkButton uses removeAttribute to remove 'href' - removed as condition from JS setAttribute.
- TActiveListBox uses removeAttribute to remove 'multiple' - removed as condition from JS setAttribute.
- TClientScriptManager:
  - _calls **`fxLoadPradoJavascript`** global event for plugins to add their own script packages._ (from ensurePradoJavascript)
  - complete docblocks.
  - static getPradoScriptsFolders; like getPradoPackages and getPradoScripts.
  - ensurePradoJavascript before use of static $_script* vars to ensure packages are loaded.
  - loadPradoJavascript to load the Prado JS packages, etc.
  - Complete and full unit tests
- TXmlTransform unit test minor update

### DIFF
--- a/framework/Web/UI/TClientScriptManager.php
+++ b/framework/Web/UI/TClientScriptManager.php
@@ -13,6 +13,7 @@ namespace Prado\Web\UI;
 
 use Prado\Prado;
 use Prado\TApplicationMode;
+use Prado\TEventResults;
 use Prado\Exceptions\TInvalidOperationException;
 use Prado\Web\Javascripts\TJavaScript;
 use Prado\Web\Javascripts\TJavaScriptAsset;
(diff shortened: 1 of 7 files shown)


framework/Web/UI/TClientScriptManager.php
```diff
@@ -221,41 +383,66 @@ protected function registerPradoScriptInternal($name)
...
 	{
+		static::ensurePradoJavascript();
+
 		if (!isset(self::$_scriptsFolders[$script])) {
 			$this->registerPradoScriptInternal($script);
 		}
 
 		$base = Prado::getPathOfNameSpace(self::$_scriptsFolders[$script]);
-		$assets = Prado::getApplication()->getAssetManager();
+		$assets = $this->getApplication()->getAssetManager();
```
ctrlaltca, line 405: Is this intentional?

framework/Web/UI/TClientScriptManager.php
```diff
@@ -221,41 +383,66 @@ protected function registerPradoScriptInternal($name)
...
```
ctrlaltca, line 429: Is this intentional, too?

framework/Web/UI/TClientScriptManager.php
```diff
@@ -266,12 +453,20 @@ public function getScriptUrls()
...
+	 *
+	 * @param string $base The base folder path (javascript or css package path).
+	 * @return array Tuple containing `[$path, $url]` where:
+	 *   - `string $path` Local filesystem path to the published asset
+	 *   - `string $url` URL to access the published asset
 	 */
 	protected function getPackagePathUrl($base)
 	{
-		$assets = Prado::getApplication()->getAssetManager();
+		$assets = $this->getApplication()->getAssetManager();
```
ctrlaltca, line 469: Is this intentional, too?

framework/Web/UI/TClientScriptManager.php
```diff
@@ -556,7 +809,7 @@ public function getStyleSheetUrls()
 			}, $this->_styleSheetFiles)
 		);
 
-		foreach (Prado::getApplication()->getAssetManager()->getPublished() as $path => $url) {
+		foreach ($this->getApplication()->getAssetManager()->getPublished() as $path => $url) {
```
ctrlaltca, line 812: Is this intentional, too?

ctrlaltca (commented):
Looks good to me, but I don't get if the changes from  `Prado::getApplication()` to `$this->getApplication()` were intentional. I'm not against them anyway.
I restarted the scrutinizer job,  it failed on fetching the postgres docker CI image

framework/Web/UI/TClientScriptManager.php
belisoful: My thought was that **TClientScriptManager is a TApplicationComponent with its own "getApplication" function**, we should probably be pulling from that rather than from the static Prado object.

I know TApplicationComponent::getApplication just pulls from Prado::getApplication.

_If/when PRADO supports multiple TApplications simultaneously, this would be a necessary change._
This is a little bit of forward thinking.  A good practice?

framework/Web/UI/TClientScriptManager.php
belisoful: see below.

framework/Web/UI/TClientScriptManager.php
belisoful: see below.

framework/Web/UI/TClientScriptManager.php
belisoful: see below.

belisoful:
I've had tests randomly fail, and not because of PRADO.  I am familiar with 2 or 3 date bugs that caused failures due to days of the month and garbage collection influencing tests (that one instance in cron). I'm not talking about those kinds of things.
The checks sometimes just _fail_.  I've had to restart the workflow tests once in a while. 2-3 times a month or so.

belisoful:
these `fx` global events are very useful for enabling (plugin extension) modules to integrate into the framework.  like the DB metadata fallback if not found as a PDO driver.

framework/Web/UI/TClientScriptManager.php
ctrlaltca: It makes sense, thank you for explaining it to me.

PR #1061, opened 2026-04-15
TSecurityManager adds EncryptionKeyAlgorithm

belisoful:
TSecurityManager provides for encrypt/decrypt.  It uses md5 to hash the EncryptionKey for the secret key in encrypt/decrypt. That may not be the easiest attack vulnerability, it's certainly far from perfect.

I'd like to change out the EncryptionKey hash algorithm for something stronger, but we need backward compatibility.

Adds methods for getting supportedHashAlgorithms and supportedCipherAlgorithms.  no more guessing about the internal black box of what algorithms are supported.

 lots of new unit tests for encrypt/decrypt.


belisoful:
BTW, I don't think we should have been setting the Algorithms before validating them.  if they fail validation, they'll be set to an invalid value if the exception is caught.

ctrlaltca (approved):
Big thumbs up

PR #1062, opened 2026-04-15
TEventParameter adds ArrayAccess to Parameter (when array)

belisoful:
Adds ParameterChanged property, and method for handlers to flag that the parameter has been changed.
Flagging the Parameter as changed by an event handler is required if the parameter is an object like TMap where changes to the parameter object are not registered as direct changes to the Event Parameter in the TEventParameter instance.
with unit test for the class

### DIFF
--- a/framework/TEventParameter.php
+++ b/framework/TEventParameter.php
@@ -14,6 +14,8 @@
 
 namespace Prado;
 
+use ArrayAccess;
+
 /**
  * TEventParameter class.
  *
@@ -24,13 +26,14 @@ namespace Prado;
  * or by constructor parameter.
  *
  * @author Qiang Xue <qiang.xue@gmail.com>
+ * @author Brad Anderson <belisoful@icloud.com>
  * @since 3.0
  */
-class TEventParameter extends \Prado\TComponent implements IEventParameter
+class TEventParameter extends \Prado\TComponent implements IEventParameter, \ArrayAccess
 {
 	private string $_eventName = '';
 	private mixed $_param;
-
+	private bool $_initialParameterNull;
 	/**
 	 * Constructor.
 	 * @param null|mixed $parameter parameter of the event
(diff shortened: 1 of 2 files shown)


framework/TEventParameter.php
```diff
@@ -77,4 +90,77 @@ public function setParameter(mixed $value)
 	{
 		$this->_param = $value;
 	}
+
+	/**
+	 * @return bool if the initial parameter was null
```
ctrlaltca: Phpdoc needs fixing

framework/TEventParameter.php
```diff
@@ -60,6 +64,15 @@ public function setEventName(string $value)
 		$this->_eventName = $value;
 	}
 
+	/**
+	 * @return bool if the initial parameter was null
+	 * @since 4.3.3
+	 */
+	public function getInitialParameterNull(): bool
```
ctrlaltca: Is this meant to be used by something external? I see no internal usage in prado.
Also,

framework/TEventParameter.php
belisoful: indeed....

framework/TEventParameter.php
belisoful: `offsetSet` has `$this->_param ??= [];` which changes the initial null parameter into an array.  This method tells callers/callees if the Parameter started as Null, but was 'filled in" with a setParameter or made an array (from null).

Maybe this should be a "getIsChanged" to flag if the Parameter has changed?  I'm leaning in this direction rather than InitialParameterNull.

belisoful: This has been updated to getParameterChanged.  Much more useful and does what I had originally "intended" for getInitialParameterNull.
I'm glad you questioned these things.  The initial idea was a reasonable starting approximation for the final code.

framework/TEventParameter.php
belisoful: fixed.

belisoful:
I like this version (with ParameterChanged) much better than the initial PR.  I hope you do too.  phpdoc has been updated.

ctrlaltca:
Yay, thank you!

PR #1065, opened 2026-04-18
Better integration of PRADO with PHPStan

belisoful:
unit tests for the PRADO PHPStan extensions via fixture
Adds the following PHPStan extensions:
- Prado::method_visible indicates the method it's checking.
- TComponent::hasMethod() indicates the method, and property when (get/set; js aware)
- TComponent:: canSetProperty() indicates a set property.
- TComponent:: canGetProperty() indicates a get property.
- MethodReflection isReadable/isWritable properties for reflection
- get/set methods are PRADO Property magic; JS aware (getjs/setjs).

### DIFF
--- a/agents/framework/PHPStan/INDEX.md
+++ b/agents/framework/PHPStan/INDEX.md
@@ -1,11 +1,11 @@
 # PHPStan/INDEX.md
 
 ### Directories
-[framework](./INDEX.md) / **`PHPStan/INDEX.md`**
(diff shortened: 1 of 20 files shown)


belisoful:
BTW, the unit test is failing because `asset-packagist.org` did not renew their domain registration!

belisoful:
At least two of my branches require these PHPStan extensions to validate properly.

PR #1066, opened 2026-04-19
TSimpleDateFormatter extended support for Java's SimpleDateFormat

belisoful:
with lots of unit tests


framework/Util/TSimpleDateFormatter.php
```diff
@@ -185,26 +309,23 @@ public function getDayMonthYearOrdering()
 	 */
 	private function getDate($value)
 	{
+		$date = new \DateTime();
```
ctrlaltca: It's not broken, but why create the $date object twice if it's not numeric?

framework/Util/TSimpleDateFormatter.php
```diff
@@ -216,200 +337,217 @@ public function isValidDate($value)
...
+		return $s->getTimestamp();
 	}
 
 	/**
-	 * Gets a portion of a string, uses iconv_substr.
-	 * @param mixed $string
-	 * @param mixed $start
-	 * @param mixed $length
-	 */
-	private function substring($string, $start, $length)
```
ctrlaltca, line 380: I guess this was meant to support multibyte languages dates (eg. japanese month names) .
I'm not sure it really worked since we had no tests for it.
It would be nice in the future to add some tests for it (no need to do it in this PR).

ctrlaltca (approved):

framework/Util/TSimpleDateFormatter.php
belisoful: Let me look into this.  I'd rather get this right without having to go back.  You make a valid point.
I was considering moving this class to the \IntlDateTime for language support.  What do you think about that shift? too?

I have the old version for the agents to compare this to.  I'll have it do another pass or ten on this and see what it can do.

framework/Util/TSimpleDateFormatter.php
belisoful: on it...

belisoful:
I've isolated the issue.

TSimpleDateFormatter is currently without multi-byte support because \DateTime only output English for names of the months, days of the week.  This doesn't mean it shouldn't "support" multi-byte" text.  It should, absolutely.  The limitation is \DateTime.

The output of TSimpleDateFormatter will always be English.  making it less useful to parse/format in multi-byte.  It's basically going half way.

I recommend we upgrade to \IntlDateTime, and add Culture (default: 'en') properties for multi-byte character output. We already have `Charset`

framework/Util/TSimpleDateFormatter.php
belisoful: BTW, it didn't support multi-byte characters before because of this:
```php
	private function length($string)
	{
		//use iconv_strlen or just strlen?
		return strlen($string);
	}
```

framework/Util/TSimpleDateFormatter.php
belisoful: reverted.

ctrlaltca:
Just wow! Great job!

PR #1068, opened 2026-04-20
TInitializedTrait to mark the initiallization of modules.

belisoful:
Converted modules from custom init marking to trait.

### DIFF
--- a/framework/Util/Traits/TInitializedTrait.php
+++ b/framework/Util/Traits/TInitializedTrait.php
@@ -18,8 +18,8 @@ use Prado\Exceptions\TInvalidOperationException;
  * TInitializedTrait provides initialization state management for classes.
  *
  * This trait tracks whether the owning class has completed its `init()` phase,
- * allowing configuration properties to remain mutable during initialization 
- * while preventing changes afterward. It provides guard methods to enforce 
+ * allowing configuration properties to remain mutable during initialization
+ * while preventing changes afterward. It provides guard methods to enforce
  * initialization requirements on properties and methods.
  *
  * Basic usage:
--- a/framework/Util/Traits/TInitializedTrait.php
+++ b/framework/Util/Traits/TInitializedTrait.php
@@ -1,0 +1,271 @@
+<?php
+
+/**
+ * TInitializedTrait class file
+ *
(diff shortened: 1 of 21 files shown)


framework/Util/Traits/TInitializedTrait.php
```diff
@@ -0,0 +1,271 @@
...
+use Prado\Data\TDataSourceConfig;
+use Prado\Data\TDbConnection;
+use Prado\Exceptions\TInvalidOperationException;
+
+/**
+ * TInitializedTrait class.
+ * ```php
+ * public function init($config): void
+ * {
+ *    $this->markInitialized();
```
ctrlaltca: This should be `$this->markStartInitialize();` i guess

framework/Util/Traits/TInitializedTrait.php
```diff
@@ -0,0 +1,271 @@
...
+ *
+ * @author Brad Anderson <belisoful@icloud.com>
+ * @link https://github.com/pradosoft/prado
+ * @license https://github.com/pradosoft/prado/blob/master/LICENSE
+ */
+
+namespace Prado\Util\Traits;
+
+use Prado\Data\TDataSourceConfig;
+use Prado\Data\TDbConnection;
```
ctrlaltca: TDataSourceConfig and TDbConnection are unused and can be removed

ctrlaltca (changes requested):
Nice! If I got it correctly `initialized_property_unchangeable` should be added to messages.txt and a lot of `*_property_unchangeable` can be removed from that same file

framework/Util/Traits/TInitializedTrait.php
belisoful: It needs more documentation.  The use of markStartInitialized requires a slightly different init process. I didn't use these the markStartInitialized in our modules because I didn't want to rock that boat.  just replacing the existing is good enough for now.  

 and I'll add the messages.

belisoful: fixed

framework/Util/Traits/TInitializedTrait.php
belisoful: fixed

ctrlaltca (approved):

PR #1071, opened 2026-04-24
TAssetManager Update - symlinking, timestamp asset urls

belisoful:
- symlink and forceCopy assets.
- asset url time stamps.
- callback for hash, beforeCopy, and afterCopy.
- assetMap.
- “only” to limit the paths of a directory copy.
- “exempt” to exclude the paths of a directory copy
- case sensitivity to the only and exempt pattern matching.
- File Mode of the copied or symlink files.
- Directory Mode of copied directories, defaults to Prado::getDefaultDirPermissions().

Unit tests


belisoful:
As much as I like this update, I am not a fan.  I've got something much better...    This PR is a compromise.

framework/Web/TAssetManager.php
```diff
@@ -265,25 +658,76 @@ protected function copyFile($src, $dst)
...
 			while ($file = @readdir($folder)) {
-				if ($file === '.' || $file === '..' || $file === '.svn' || $file === '.git') {
+				if ($file === '.' || $file === '..') {
 					continue;
-				} elseif (is_file($src . DIRECTORY_SEPARATOR . $file)) {
-					if (@filemtime($dst . DIRECTORY_SEPARATOR . $file) < @filemtime($src . DIRECTORY_SEPARATOR . $file)) {
-						@copy($src . DIRECTORY_SEPARATOR . $file, $dst . DIRECTORY_SEPARATOR . $file);
-						@chmod($dst . DIRECTORY_SEPARATOR . $file, Prado::getDefaultFilePermissions());
+				}
+				if ($file[0] === '.' && ($file === '.svn' || $file === '.git' || $file === '.github' || $file === '.claude' || strlen($file) > 1)) {
```
ctrlaltca, line 690: Should we put these in a static "blacklist" array so that they are more explicit?

ctrlaltca (approved):

framework/Web/TAssetManager.php
belisoful: 72ea92afdba7dac93534ce8d53263030f8e3b886 fixes this.

PR #1073, opened 2026-04-24
TEnumerable update

belisoful:
adds static methods to TEnumerable:
- valueOf gets the value of a Constant, with case insensitive option.
- hasConstant checks if a constant exists, with case insensitive option.
- constantOf gets the constant of a given value, with case insensitive option.

These methods provide for working with TEnumerable constants from a variable, like a property. 

Caching the TEnumerable Reflection class.

### DIFF
--- a/framework/Web/UI/TWebColors.php
+++ b/framework/Web/UI/TWebColors.php
@@ -19,7 +19,6 @@ namespace Prado\Web\UI;
  * @author Brad Anderson <belisoful@icloud.com>
  * @since 4.3.0
  * @see https://en.wikipedia.org/wiki/Web_colors  Web Colors
- * @todo 4.4 TWebColors => TWebColor, singular - not plural #1072
  */
 class TWebColors extends \Prado\TEnumerable
 {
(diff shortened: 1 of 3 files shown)


framework/Web/UI/TWebColors.php
```diff
@@ -19,6 +19,7 @@
  * @author Brad Anderson <belisoful@icloud.com>
  * @since 4.3.0
  * @see https://en.wikipedia.org/wiki/Web_colors  Web Colors
+ * @todo 4.4 TWebColors => TWebColor, singular - not plural #1072
```
ctrlaltca: we can fix it now and add a
```php
/*
 * @todo 4.4 remove compatibility alias
 */
class_alias('TWebColor', 'TWebColors');
```
No need to do it in this PR, as you prefer

ctrlaltca (approved):

framework/Web/UI/TWebColors.php
belisoful: Even better.  That's so cool.  The only place where class_alias is used is in `Prado.php` to alias the Prado class. I'll put this in a different PR.

ctrlaltca: You could just put in inside the same file at the end of the class, but this would probably break autoloading using classes.php.
Maybe create an empty TWebColors.php with just an include and the class_alias?

PR #1078, opened 2026-04-29
TModuleView & TControl::createControlChildren parent chain (for attached behaviors)

belisoful:
TModuleView - only instancing children on presence of module, fallback template
TControl::createControlChildren has a parent chain for triggering attached behavior actions on the same
TDataSize uses TI18NControlTrait

Use case 1 for TModuleView.  Developer installs a PRADO 4 extension plugin that implements several template controls that are used, but the developer doesn't want the page to break if the extension is un-installed.

Use case2 for TModuleView: Plugin Developers can code their own pages and (template) controls to key on other installed modules/controls; recursive plugin behavior for plugins.


ctrlaltca:
If you can please try to split these PR into functional units, they are quite hard to review all-in-one.
Eg. the TDataSize changes could be a different PR.

PR #1087, opened 2026-05-04
CSP - merging latest THttpResponse

belisoful:
this bring prado:csp up to date with HEAD.
The only file that actually matters is THttpResponse.


belisoful:
There are some critical bugs but I'd like to update this to HEAD first.

There is one element of the CSP that I think we should probably implement "report-url" and "report-to".  This requires something like a TCSPReporterService which is deployed when the report feature is on.  The THttpHeaderCSP would inject a second header into the manager to trigger reporting and where.  The URL to report to can be an option, or automatic (with the TCSPReporterService).

The TCSPReporterService would ideally have a Prado::log and trigger a service "on" event.

ctrlaltca:
I've rebased the csp branch to master, so this is no more needed.

PR #1088, opened 2026-05-04
Database Modernization (insertOrIgnore, Upsert, Interfaces, PDO Db Capabilities [centralization], unit tests)

belisoful:
Here is the list of changes to Data:
- TDbDriverCapabilities - centralized location of all Driver specific implementations.
   - Adds fxActiveRecordScaffoldInputClass for getting non-PDO driver classes.
   - Adds fxDataGetMetaDataClass for getting non-PDO driver classes.
- Insert Or Ignore in Command Builder.
- Upsert in Command Builder.
- InsertOrIgnore and Upsert for ActiveRecord.
- InsertOrIgnore and Upsert for SqlMap.
- `TDbDriver` class to centralize the names of the PDO Drivers.
- `TDataCharset` class to normalize Data Charsets.
- Db Modules with PDO Driver Names updated to use TDbDriver for Driver Name.
- Fixed processException in PradoUnit (a class for unit testing and suppressing multiple errors, eg Db connection).
       PradoUnit - [@todo - this feature should be abstracted to be key based and allow other unit tests to not repeat multiple of the same errors]
- tweaks Github CI to remove pdo_firebird and pdo_sqlsrv from the main PHP CI matrix so those tests are skipped.
       (these are on by default. are they really so common?)
- TDbTransaction::beginTransaction for "serial" transactions. Invalidated by TDbConnection::beginTransaction, which emits a new (replacement) TDbTransaction.
- Lots of unit tests for classes in Data hierarchy.
- Lots of unit tests for DbSpecific.  Validation of each database system has exposed several bugs, all fixed.
- Better class doc blocks for Data classes.  This has been missing for far too long.
- Adds TDbConnection methods:
  - commit/rollback. transactions are features of the connection; forwarded to the current transaction.
  - getLastTransaction. returns what is in the in the connection transaction property regardless of active. TDbTransaction::beginTransaction uses it to ensure that it is still the operational transaction for the connection; meaning, something else called TDbConnection::beginTransaction before the old transaction tried to beginTransaction.
  - __sleep => _getZappableSleepProps.
  - extractDriverFromDsn.
  - extractCharsetFromDsn.
  - applyCharsetToDsn.
  - getDatabaseCharset (gets the charset from database connection that it can be retrieved).
  - createTransaction (protected).
  - getHasAutoCommit.
  - assertActive.
- Changed Mssql to SqlSrv to match the driver.  "mssql" is not a PDO extension, it's a PHP extension.  This is to make room for the possibility of implementing mssql PHP extension rather than just PDO.
- TTableGateway::getTableExists to check for the existence of a table.

About 70% of this PR is unit tests.

Tested but removed:
- a "TDbSerialTransaction" was tested, but conflated implicit transactions and explicit transactions making it simply an invalid ideation.  I learned about the implicit and explicit state of transactions in PDO for AutoCommit, and, separately, of Firebird always being in a transaction on a connection.  This was removed and replaced by TDbTransaction::beginTransaction.


belisoful:
Quick update - 
This is coming along nicely,  I like where it is.  The DB unit testing we had was really hit or miss.  The difficulty is not so much in the code, but organizing the Database unit tests themselves.

The Testing of the entire DB system and all supported PDO drivers is big; huge.

That's why I'm breaking it apart into a more fine grain than we currently have.  Isolation, Encapsulation.

Data Testing Infrastructure needs improvements too.  The local tests don't run the initdb scripts.   I'm looking into that.

PR #1095, opened 2026-05-08
Adds TApplication::onConfigurationComplete, getServiceIdByClass, and getServiceIdsByClass

belisoful:
TCspReporterService might get installed programmatically after modules are configured but before service discovery by THttpHeadersManager.

CSP branch needs to be reset for this change, to support the CSP Reporter Service.


ctrlaltca:
> CSP branch needs to be reset for this change, to support the CSP Reporter Service.

Done

PR #1096, opened 2026-05-08
Vitest unit tests for Prado JS - new Github CI job

belisoful:
- adds new JS CI workflow to test Prado Javascript.
- .gitignore of Node.js folders/files.
- composer.json adds js testing scripts - and fills out scripts for csfixer, and phpstan..
- `eslint.config.js` for linting javascript. (js equivalent of php-cs-fixer).
- `vitest.config.js` for testing javascript.
- adds `package.json` for Node.js - Supplements PHP `composer.json`. (those files do the same thing).

- Vitest unit tests for all the Prado Javascript.
- Minor Prado JS bug fixes.- 

**Notes:**
JS unit tests can easily be run locally.  No more guessing or live testing for Prado JS validation.

_vitest_ was selected over jest or other js test platforms because it has more modern support that eventually PRADO should move to (in regards to ESM and ES6).  It should support the current (old school) JS model and its conversion to the new school JS models.


belisoful:
As an aside, @ctrlaltca Selenium tests are great but older solution, what are your thoughts on integrating Playwright or replacing/upgrading selenium tests with Playwright?  Playwright could provide real functional integration tests in actual multiple different browsers like Chrome, the current M$ Edge, Safari.  [at least what I know of it].

ctrlaltca:
Yeah Playwright is cool! Prado tests were historically written with old Selenium-RC, and later ported to Selenium webdriver. The amount of job required to port them again to Playwright is probably massive, but if you feel like you want to do it, go for it!

belisoful:
I thought doing a full suite of JS unit tests was massive. It occurred to me about 3 days ago.

And thank you for the blessing for reworking the functional tests into playwright.  It's not high priority, but it'll be on the list for sure.

PR #1100, opened 2026-05-12
TControl Render Filter - TRenderFilterParameter

belisoful:
These two parts are very simple and integrate where they touch. This is a part of the larger revamping of Web.UI and WebControls

One part is putting the "Adapter or Control" into a a method because it was repeated so much.  This basically necessitated an Interface unifying the TControlAdapter and TControl.  This is good form and how to "combine" them.

The over all sense in TControl is so much cleaner with getAdapterControl.

The Second part is onRenderFilter.  `renderControl` and `renderChildren` insert a new writer, renders, reads the writer, restores the old writer, filters the text by attached handlers, and then writes the result to the restored writer.

The TRenderFilterParameter is interesting because it provides DOM conversion, both ways, and a DOM walk.  The elements can be walked individually as well.

This is where the IEventCycleParameter came from: for conversion of the DOM back to HTML at the end of the event in a formal way.


belisoful:
The two parts were developed together, btw.

belisoful:
BTW, the Prado-demos PR is to allow Prado-demos to be included as a dev dependency....  for playwright.

also I have a refactored TApplication, TShellApplication, a new TTestApplication, and a few other new features to PradoUnit.  This came about because of the Service interface that we need in TApplication for THttpHeadersManager.   After having a look at the Services API, it seemed like there were a few other changes that were needed to move into the future-present.

The TApplication refactor will be needed for reset of `csp` to be able to use the Service API over there.    I'm finalizing The TApplication Refactor, then I'll be able to have another look at THttpHeadersManager, which is looking pretty good.

PR #1105, opened 2026-05-13
TApplication Refactor - TShellApplication, TTestApplication; PradoUnit upgrade

belisoful:
TApplication changes:
 - Corrects the php doc application lifecycle order.
 - Functional abstraction for fine grain subclass control.
 - Formalized the creation of default object instances, like TSecurityManager, THttpResponse, TTemplateManager, etc.
 - onGlobalStateChange to inform event handlers
 - `onConfigurationComplete` changed to just `onConfiguration` 
 - Dividing into Sections by API
 - Formalizing LazyModule API
 - Filling out Services API
   - registerService validates class as guard, not at startService time.
 - `configFileName` uses new app property rather than static.
 - Reorganized some properties
 - moved Configuration loading of Services to after App Properties but before App Parameters; services are data only types 
 - Centralizes UniqueID hashing into `generateAppUniqueId` - to change to sha1 in Prado v4.4
 - Abstracting Application Steps and step index.
 - registerApplication so TestApplications can trap before setting the global property.
 - Constants key from static.
 - New Exceptions messages.txt
 - Unit Tests

TShellApplication changes:
- Supports cli without a repo, uses repository `./.runtime/` or sys_tmp for a semi-stable last resort.
- Enables cache and active record only when configured.
- Provides accessors to its properties.
- Functional abstraction for fine grain subclass control.
- Move `processArguments` from [onInitComplete, 20] to [onConfiguration, 20]
- Unit Tests

TTestApplicationTrait - common features of `TTestApplication ` and `TTestShellApplication `
TTestApplication - easier way to create an app in all unit tests. 
 - This makes creating new TApplication in a test environment much easier.
TTestShellApplication  -  easier way to create a shell app unit test.
 - Formal (meta) unit tests

PradoUnit changes:
 - Formalizing the class.
 - Snapshot and restore the state of an object.
 - Documentation.
 - skipSlowTests - checks environment variable for `PRADO_UNITTEST_SKIP_SLOW`.
 - IsGitHubActions - checks if the context is a GitHub Action.
 - getCIEnvironment - checks for common CI environment variables.
 - Formal (meta) unit tests.


belisoful:
@ctrlaltca  _This_ is the TApplication Services API that THttpHeadersManager requires.
   and prado-demos, vendor branch had one too many `../`.   PR is pending.   I've been testing the change, so it certainly works.

framework/TApplication.php
```diff
@@ -449,27 +575,61 @@ public function getRequestCompleted()
...
+				$this->_globals[$key] = $value;
+				$this->_stateChanged = true;
+				$changed = true;
+			}
+		}
+
+		if ($changed) {
+			$payload = ['key' => $key, 'value' => $value, 'isDefault' => $isDefault, 'isNew' => !$isset];
+			if ($isset) {
+				$payload['oldValue'] = $oldValue;
```
ctrlaltca: `$oldValue` is not be defined in all code paths

framework/TApplication.php
```diff
@@ -479,12 +639,57 @@ public function setGlobalState($key, $value, $defaultValue = null, $forceSave =
...
-		$this->_stateChanged = true;
-		unset($this->_globals[$key]);
+		if (array_key_exists($key, $this->_globals)) {
+			$oldValue = $this->_globals[$key];
+			unset($this->_globals[$key]);
+			$this->_stateChanged = true;
+			$this->onGlobalStateChange(new TEventParameter([
+				'key' => $key,
+				'isUnset' => true,
+				'oldValue' => $oldValue,
```
ctrlaltca: The available keys for this parameter (key, isUnset, oldValue) are different than the ones defined by `setGlobalState` (key, value, isDefault, isNew, ?oldValue).
When i receive the event, some of them may not exists, so i'm supposed to always check for any possible combination.
Could we ensure oldValue always exists and condensate isUnset, isDefault, isNew in a single "flags" field? So that we have 4 keys that are always the same and all defined.

framework/TApplication.php
```diff
@@ -687,34 +1082,124 @@ public function setService($value)
...
+	 * The returned array is a three-element tuple `[$class, $properties, $config]`:
+	 * - `$class` (`string`) — fully-qualified class name of the service.
+	 * - `$properties` (`array`) — name → value map of initial properties applied before `init()`.
+	 * - `$config` (`null|array|TXmlElement`) — optional extra configuration element.
+	 *
+	 * @param string $id service ID to look for.
+	 * @return ?array the three-element registry tuple, or `null` if the ID is not registered.
+	 * @since 4.3.3
+	 */
+	public function getServiceId(string $id): ?array
```
ctrlaltca: getServiceById?

framework/TApplication.php
```diff
@@ -687,34 +1082,124 @@ public function setService($value)
...
+	}
+
+	/**
+	 * Returns the full service registry as an array keyed by service ID.
+	 * Each value is a three-element tuple `[$class, $properties, $config]` as
+	 * described in {@see getServiceId()}.
+	 * @return array<string, array> the registered service configurations.
+	 * @since 4.3.3
+	 */
+	public function getServiceIds(): array
```
ctrlaltca: getServices?

belisoful:
Given the importance of TApplication, this has been "proofed" dozens of times.  Every aspect has been pulled apart for proper abstraction.  And that is made use of in TShellApplication, and a new testing infrastructure with TTestApplication for testing with a Test Application.

framework/TApplication.php
```diff
@@ -1133,31 +1918,36 @@ public function applyConfiguration($config, $withinService = false)
...
+		// load services, provide for modules
+		foreach ($config->getServices() as $serviceID => $serviceConfig) {
+			$this->registerService($serviceID, ...$serviceConfig);
 		}
 
+		// Default Page Service is registered at construct,
+		//		setPageServiceId changes the registered service within _services
+
 		// load parameters
+		$appParams = $this->getParameters();
```
ctrlaltca, line 1949: the "return by reference"-style access of TMap tricked me here. Just a note to my future self, nothing needs to be fixed

framework/TApplication.php
```diff
@@ -63,27 +65,35 @@
...
- * - flushOutput : the application flushes output to client side.
- * - onEndRequest : this is the last stage a request is being completed
- * - [destruct] : destruction of the application instance
+ * - [construct] : The application instance has been constructed.
+ * - [initApplication] : Configuration has been loaded; modules and the requested service have been instantiated.
+ * - onConfiguration : Configuration has been fully applied. The request has not yet been resolved and no service has been started.
+ * - onInitComplete : The service has been initialized. For {@see Shell\TShellApplication}, argument processing is attached to this event.
+ * - [run] : The primary request lifecycle has begun.
+ * - onBeginRequest : Application initialization has completed.
+ * - onLoadState : State loading has begun.
```
ctrlaltca, line 84: ~I need to think about possible implications of moving onLoad before onAuth*~
EDIT: nevermind, my fault for reviewing before coffe

framework/TApplication.php
belisoful: the problem is if the oldValue has a value of `null`. This is an edge case where the defaultValue is not `null` so there can be a value of `null`.

I love the idea of a flags.  This is supposed to be a parameter class (`TArrayValueChangeParameter`) by itself at some point.  Your suggestions are compatible with that co-creation in v4.4.  My thought process was that it could be used for changes to Maps, and Collections at large as well.

framework/TApplication.php
belisoful: There are two parallel concepts. One is the Instanced Service and the other is the registered Service Ids and object definitions.

I had considered `getServiceById` but that sounds too much like instancing a Service due to `getService`.

So the delineation is between _Service and ServiceId_ being the _Instanced service vs possible registered services_.

ctrlaltca: Got it, thank you

framework/TApplication.php
belisoful: This has the same issue of being so similar to getService, and has no implied distinction that it's not fully instanced versions of all the services.

We could rename it to ServiceClasses to make the distinction.  But I found that to be I bit cumbersome in length.

ServiceDefs, RegisteredServices?  I'd be ok with RegisteredServices despite the length.

framework/TApplication.php
belisoful: I ran into the same issue when I started analyzing.

framework/TApplication.php
belisoful: There wouldn't be a distinction between the lack of an oldValue an oldValue as null.  See the following comment on the same...

belisoful: maybe this could be a flag too.

ctrlaltca: The isNew flag would suffice this case

belisoful: I'm going to just put together the class for Global State Change Parameter.   It'll have the "is*" as flags and as getters/setters.   It'll be useful in the future for Collections when that changes are passed to event handlers, hopefully in v4.4.

ctrlaltca: Given the massive changes, it surely won't be a minor version. Feel free to target version 4.4 and break what's needed.

framework/TApplication.php
ctrlaltca: getServiceRegistry, getRegisteredServices, ... whatever you prefer

framework/TApplication.php
belisoful: I've added the TCollectionItemChangeParameter for onGlobalStateChange.  I'm hoping have it more involved in actual collections in v4.4. but I like what it is.  Just changing the names of the ServiceId to indicate the "registry" part and we'll have a handle on this PR.

framework/TApplication.php
belisoful: ok.  the original idea was ServiceIds, but I see how that can be confusing in regards to the instanced service.  

I'll tweak the names on these.

belisoful:
regarding "* - onLoadState : State loading has begun." before onAuth.  I was documenting the steps as they were.  The documentation had incorrect steps.

The steps before and still currently are:
```php
	private static $_steps = [
		'onBeginRequest',
		'onLoadState',               // <- onLoadState
		'onLoadStateComplete',
		'onAuthentication',
		'onAuthenticationComplete',
		'onAuthorization',
		'onAuthorizationComplete',
		'onPreRunService',
		'runService',
		'onSaveState',
		'onSaveStateComplete',
		'onPreFlushOutput',
		'flushOutput',
	];
```
The problem is what the documentation said:
```markdown
  - onBeginRequest : this event happens right after application initialization
  - onAuthentication : this event happens when authentication is needed for the current request
  - onAuthenticationComplete : this event happens right after the authentication is done for the current request
  - onAuthorization : this event happens when authorization is needed for the current request
  - onAuthorizationComplete : this event happens right after the authorization is done for the current request
  _- **onLoadState** : this event happens when application state needs to be loaded_
  - onLoadStateComplete : this event happens right after the application state is loaded
  - onPreRunService : this event happens right before the requested service is to run
  - runService : the requested service runs
  - onSaveState : this event happens when application needs to save its state
  - onSaveStateComplete : this event happens right after the application saves its state
  - onPreFlushOutput : this event happens right before the application flushes output to client side.
  - flushOutput : the application flushes output to client side.
  - onEndRequest : this is the last stage a request is being completed
  - [destruct] : destruction of the application instance
```

I went by the code as the "source of truth."  This is why the class doc blocks were updated.  If there are security concerns, or if we should consider moving onLoadState to after `onAuth` please let us know.

framework/TApplication.php
ctrlaltca: Thank you!

PR #1107, opened 2026-05-14
1:1 Playwright functional tests 

belisoful:
Summary:
 - 1:1 replication of selenium functional tests in playwright.
 - playwright is a Node.js package.
 - Functional tests in playwright easily run locally.
 - Chromium, Firefox, and Webkit.
 - Test Servers are automatically spun up and shut down (even in odd OS Process conditions).
 - Selenium Functional tests remain, and are much faster.
 - updates the README.md Test section
 - new Composer scripts for playwright and tweaked the existing ones for alignment.
 - Playwright has its own CI Job (like Windows and Firebird)
 - Prado-demo goes into `vendor/pradosoft/`
 - Prado-demo self downloads and updates (It's vendor/pradosoft/ is linked back to the project to test it)
     _we don't know the environment of a developer and cannot assume the directory above the project is "safe"_
- Functional Test PHP Webserver runs on 8037

- Single run CI Jobs get their PHP_VERSION from the env variable.

It turns out that the autoload.php edit to `prado-demos` wasn't needed, but is useful for projects.  Just not prado due to circularity.  Which, I guess I needed to test to realize.  It's fine.  It's a reasonable edit anyway.


drigolin:
Why not use playwright in PHP 
https://playwright-php.dev/

belisoful:
Moving to PHP should be easy now that it's already there.   I was thinking of functional testing taking place in javascript, especially after the recent JS vitest unit tests.  Point taken.

belisoful:
I looked into playwright-php.

playwright-php is a php wrapper for Node.js playwright
playwright-php appears to be more for projects that don't have Node.js.
playwright-php does not have video recording per test like the OG.
    This is a vital feature.  And I suspect that it will be needed as we start adding HTML 5 ActiveControls.

We already have Node.js integrated into Prado for the JS eslint + vitest unit tests. This makes a wrapper redundant. 

I think this is acceptable.  Node is huge, we got it.

belisoful:
I see the webkit error.  I'm on it...

Issue #1111, opened 2026-05-15
#1078 breaks TTemplateControlInheritable


ctrlaltca:
I have a custom component containing a TDataGrid containing a TDropDownList.
The custom component inherits TTemplateControlInheritable.
Since #1078 this causes the following error:

<img width="628" height="476" alt="Image" src="https://github.com/user-attachments/assets/3d7668a8-cbd9-4608-8525-b8f1d45d909a" />

Changing the custom component to inherit TTemplateControl fixes the error.
I'm building a small testcase.

ctrlaltca:
The problem is due to the fact that TTemplateControlInheritable gets rendered twice.
Following is a small testcase based on prado-demos/helloworld.

Home.page:
```html
<html>
<body>
<com:TForm>
<com:CustomControl />
</com:TForm>
</body>
</html>
```
Home.php:
```php
<?php
Prado::using('Application.pages.CustomControl');
class Home extends TPage { }
```
CustomControl.page:
```html
Hello world<br/>
```
CustomControl.php:
```php
<?php
class CustomControl extends TTemplateControlInheritable { }
```

The resulting page:

<img width="137" height="63" alt="Image" src="https://github.com/user-attachments/assets/3d146ce6-29b8-4d06-960b-e6fd399ea2c8" />

ctrlaltca:
This is probably caused by this single addition:
https://github.com/pradosoft/prado/commit/381eb27bb215d845016b2637bcbd015fd4ddd71c#diff-40bb37e3978acd3a20390684e03ab5f3838efc89f87af5bc9c628a4c342ba15eR56

<img width="605" height="317" alt="Image" src="https://github.com/user-attachments/assets/dd2ffd32-3884-4c06-9e3b-e3b228aea5ae" />

Removing that single line restores functionality.

belisoful:
on it...

belisoful:
The parent was chained to link in the dynamic event at the root method.  There was no unit test that checked for this. I'll see if there is a more general unit test, like checking for duplication that will pick this up but also test for other code that might do the same.

Currently fixing the prado3 namespace classes bug in a more general way...

PR #1113, opened 2026-05-17
PHP FQN corrections

belisoful:
Audit of classes that are not using PHP FQN.
- TResultProperty::getPropertyValueType checks for TList.
- TComponentReflection::reflect checks against TComponent (unit tests)
- TTemplateControlInheritable::doCreateChildControlsFor checks parent for TTemplateControl.
- TDataList::renderItem checks for TDataListItem. (unit tests)


belisoful:
I found these while doing a proper fix for #1111.  The fix for TTemplateControlInheritable includes its own unit test, which tests for these commits.

PR #1120, opened 2026-05-19
PradoTestListener - start/stop Selenium Driver on proper test suite

belisoful:
Not sure if this is even needed given that the selenium tests appear to be working.  Playwright should be concurrent to selenium until v4.4, when selenium is removed.

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -26,5 +26,7 @@
     <ini name="max_execution_time" value="1200"/>
     <ini name="memory_limit" value="1G"/>
     <const name="PRADO_TEST_RUN" value="true"/>
+    <!-- indicate Selenium test suite to start/stop Driver -->
+    <const name="SELENIUM_TESTSUITE_NAME" value="selenium"/>
   </php>
 </phpunit>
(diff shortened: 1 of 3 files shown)


ctrlaltca:
Something is wrong:
```
starting PHP server on port 8037...
Starting geckodriver on port 4444...
[Tue May 19 05:34:22 2026] PHP Warning:  JIT is incompatible with third party extensions that override zend_execute_ex(). JIT disabled. in Unknown on line 0
[Tue May 19 05:34:22 2026] PHP 8.3.31 Development Server (http://127.0.0.1:8037/) started
PHPUnit 10.5.63 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.3.31
Configuration: /home/runner/work/prado/prado/phpunit.xml

No tests executed!
```

belisoful:
Is it not the same between Github CI and local?

I was doing a clean-up job to ensure that it's fully converted.  I was relying upon the Github CI to verify this change.  What is the difference?

It seems like a reasonable change that wouldn't break such things.

ctrlaltca:
https://github.com/pradosoft/prado/actions/runs/25996509888 was the last successful run for the selenium test cases
https://github.com/pradosoft/prado/actions/runs/25996601049/job/76412086145 is the first run indicating no tests.
So i guess it was #1107

belisoful:
The run-selenium-test.sh was running the old "functional" tests and getting nothing because that was updated to a different name.

Fixed.

PR #1121, opened 2026-05-19
TPgsqlMetaData - fix for null string passed into substr - PHP8 changed its behavior

belisoful:
Found this in the Data refactor.


belisoful:
Side-note: in all of this, I'm also working on the underlying "feel" of Prado.  It's hard to qualify, being intuitive would be a good start.

PR #1122, opened 2026-05-19
TErrorHandler update - THttpException and TExitException

belisoful:
THttpException - conforms to Uniform Access Principle - Self Encapsulation (UAP-SE)
TExitException - conforms to Uniform Access Principle - Self Encapsulation (UAP-SE)
TErrorHandler changes
 - uses `static::` rather than `self` in proper places.
 - updated documentation (including xml and php examples).
 - caching of privatePathReplacements.
 - recursive error handling in class, not static.
 - better date-time handling.
 - encapsulate PHP functions for unit testing (like `$_SERVER[$key]` accessed by `$this->serverGlobal($key)`).
 - conforms to UAP-SE.
 - return types for PHP 8.

### DIFF
--- a/framework/Exceptions/TErrorHandler.php
+++ b/framework/Exceptions/TErrorHandler.php
@@ -92,10 +92,14 @@ class TErrorHandler extends \Prado\TModule
 
 	/**
 	 * @var ?array cached path replacement map for hidePrivatePathParts()
+	 * @since 4.3.3
 	 */
 	private $_privatePathReplacements;
 
-	/** @var bool whether an error is currently being handled (prevents infinite loops) */
+	/**
+	 * @var bool whether an error is currently being handled (prevents infinite loops)
+	 * @since 4.3.3
+	 */
 	private bool $_handling = false;
 
 	/**
@@ -115,10 +119,7 @@ class TErrorHandler extends \Prado\TModule
 	 */
 	protected function setAppErrorHandler()
 	{
-		if (!($app = $this->getApplication())) {
-			return;
-		}
-		$app->setErrorHandler($this);
+		$this->getApplication()?->setErrorHandler($this);
 	}
 
 	/**
--- a/framework/Exceptions/TErrorHandler.php
+++ b/framework/Exceptions/TErrorHandler.php
@@ -18,34 +18,50 @@ use Prado\Prado;
  *
  * TErrorHandler handles all PHP user errors and exceptions generated during
  * servicing user requests. It displays these errors using different templates
- * and if possible, using languages preferred by the client user.
- * Note, PHP parsing errors cannot be caught and handled by TErrorHandler.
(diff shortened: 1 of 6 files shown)


framework/Exceptions/TErrorHandler.php
```diff
@@ -155,20 +292,21 @@ public function handleError($sender, $param)
...
+	 * Replaces the document root, the framework directory, and any per-frame source
+	 * directories found in the exception trace with safe placeholder tokens.
 	 * @param string $value
-	 * @param null|\Exception $exception
-	 * @return string
+	 * @param ?\Exception $exception
 	 * @since 3.1.6
 	 */
-	protected static function hideSecurityRelated($value, $exception = null)
+	protected function hideSecurityRelated($value, $exception = null): string
```
ctrlaltca, line 303: This deserves a mention in UPGRADE.txt
At least from my experience it's quite common to subclass TErroHandler to implement handling of custom  app exceptions

ctrlaltca (approved):

framework/Exceptions/TErrorHandler.php
```diff
@@ -191,38 +332,38 @@ function_exists('xdebug_get_function_stack')) {
...
-	 * @param int $statusCode response status code
-	 * @param \Exception $exception exception instance
+	 * Renders a user-facing error page for the given HTTP status code.
+	 * Logs non-HTTP exceptions via {@see errorLog()}.
+	 * In non-Debug mode, strips sensitive path information from the error message.
+	 * @param int $statusCode
+	 * @param \Exception $exception
 	 */
-	protected function handleExternalError($statusCode, $exception)
+	protected function handleExternalError($statusCode, $exception): void
```
ctrlaltca, line 341: This one deserves a mention in UPGRADE.txt, too:
```
Declaration of MyCustomErrorHandler::handleExternalError($statusCode, $exception) must be compatible with Prado\Exceptions\[TErrorHandler](https://pradosoft.github.io/docs/manual/class-Prado.Exceptions.TErrorHandler.html)::handleExternalError($statusCode, $exception): void
```

Note: adding the `: void` return type hint to the custom error handler makes it compatible both with the old and the new method signature

framework/Exceptions/TErrorHandler.php
belisoful: Let me fix this....  It'll be both, with the main through the static.

framework/Exceptions/TErrorHandler.php
belisoful: Let me fix this too.There may be several places to revert the type

belisoful:
That change was in my blind spot apparently.  I did verify that the parameters were the same.  I'll get the method parameter and return type back.  it should be the same as 4.3.2.  100%.

PR #1124, opened 2026-05-20
TErrorHandler correction for prior methods staying the same.

belisoful:
My apologies for mucking this.  I'm glad you called this error out @ctrlaltca.  My intention was to keep it the same in 4.3.3.

For v4.4 we should embrace method parameter and return  types.  I think we can do the whole project pretty easily with these automated "AI" tools.  

I think we can do a conversion of the value based on the setter parameter type in TComponent::setSubProperty. via Reflection.  This is the method everything is routed through (though I'll need to do an audit to be sure).  TTemplate was one of the few places that it was not centralized.

We can also discuss doing a "return $this;" at the end of property setters for chaining.  Again, I think we can just convert the whole project with the AI tools.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,4 +1,4 @@
-# Upgrading Instructions for PRADO Framework v4.3.2
+# Upgrading Instructions for PRADO Framework v4.3.3
 
 ### !!!IMPORTANT!!!
 
@@ -6,6 +6,10 @@ The following upgrading instructions are cumulative. That is,
 if you want to upgrade from version A to version C and there is
 version B between A and C, you need to following the instructions
 for both A and B.
+
+Upgrading from v4.3.2
+---------------------
+- Prado\TEventParameter adds a second constructor $readOnly
 
 Upgrading from v4.2.2
 ---------------------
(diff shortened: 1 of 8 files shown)


ctrlaltca:
You don't need to worry.
Imho we can just target the 4.4 version and start cleaning up the compatibility layers.
There's no need for a 4.3.3

belisoful:
I was really REALLY _**REALLY**_ hoping you'd say that.  I started a 4.4 branch on my fork.  It's already got a bunch of stuff ready.  I have a few more commits to make on master before they'll start validating (they are a bit advanced for where we are on master).

There will be some changes that I really think you'll like.

**Given the Security holes that were patched, It would be wise to merge _this_ PR for backward compatibility. There should only be a few other minor merges for 4.3.3 and then tag it for 4.3.3 release**. 

We should recommend upgrading to 4.3.3 for security reasons.

This will at marginally help in developing composer plugins.

The additional feature upgrades (TStyle, TTemplate) and bug fixes are good for 4.3.3.  People can at least add style fields to template web controls directly.  There are a few other changes I'd like to make to 4.3.3 before it gets tagged.  For instance, there are a few minor edits to the AGENTS.md based upon my recent testing

Recent changes just needs to be documented in HISTORY.md. IT's been a bit neglected, but it's super easy to make a summary of the git commits since 4.3.2.  When things are ready here,  I'll have it ready in a jiffy.

Then shift to 4.4 mode.

ctrlaltca:
I've merged the PR and created a new branch prado-4.3.
I'll test it for a while before tagging and releasing v4.3.3.
In the meanwhile, you can consider the master branch ready to accept changes for 4.4

PR #1125, opened 2026-05-21
Final Revisions on 4.3.3 - Data\* alignment to 4.4, TService::getInstance, and agents update

belisoful:
This is basically the last set of revisions for 4.3.3 before the closing set.

- TService::getInstance filters TApplication::getService for the static class being called.  very useful.
- TApplication::registerService fix for null or blank class names.
- Data\* alignment with Prado 4.4 regarding Interfaces, and global events (and their data).  Needed for 3rd party databases that is forward compatible.
- AGENTS.md and CLAUDE.md needed edits based on experience with the outcome of these files.
- Adds missing classes.php.
- adds @template to the doc blocks on methods that need it for phpstan.
- TConstantReflectionTrait strlen === 0 bug fix.
- TDataSourceConfig handles null.
- TDirectoryCacheDependency Bug fix: subdirectory timestamps merged properly.
- Adds TDbDriver and TDataCharset (IANA).

These changes are organized by commit.  if better resolution than looking at the whole thing is needed, looking at commits is the way.

Here is what is left on v4.3.3:
 - Update the Working Knowledge with _this_ commit.
 - Update the HISTORY.md
 - Update the version numbers.

The Working Knowledge is now current.  it's waiting on this merge to have the last update.
The HISTORY.md is looking pretty fancy for having just released 4.3.2.
Lastly, there is a script for updating all the version numbers waiting in the dev-tools right now.  you can add `--dry-run` and it'll tell you what it's updating and to what.

I don't want to do much on `master` until this is wrapped up.  Nearly there.

And FYI, I did a few audits to ensure that the (`prado-mongodb` repo)[https://github.com/belisoful/prado-mongodb] will work in the `prado-4.3` branch.  This helps ensure that we can support other databases.  One just happens to be a side project.  I'm using it to validate Prado and the interface for 3rd parties.  It won't technically start validating until 4.3.3 is tagged and released.  but it's "mostly done"?


belisoful:
open/close/etc. I'll just submit a PR on both branches.

belisoful:
Putting the final touches on 4.3.3...

PR #1127, opened 2026-05-22
Module Dependencies - Some load after others

belisoful:
Prado has always had to order the modules _by hand_.  Any modules that depend on other modules to be initialized had to be strictly ordered in the configuration.

We handled this by initializing the modules in phases.  Dependencies only works when there are "post initialization" processes, eg executed in onConfiguration, onInitComplete, or onBeginRequest.

`IModuleDependency` can be implemented by a class to tell the application what modules it depends upon to ensure they are loaded first.  This allows for Prado to configure an application without concern for module load order during initialization.

### DIFF
--- a/framework/Exceptions/messages/messages.txt
+++ b/framework/Exceptions/messages/messages.txt
@@ -68,6 +68,8 @@ application_unavailable					= Application is unavailable at this time.
 application_service_unavailable			= Service '{0}' is unavailable at this time.
 application_moduleid_duplicated			= Application module ID '{0}' is not unique.
 application_runtimepath_failed			= Unable to create runtime path '{0}'. Make sure the parent directory exists and is writable by the Web process.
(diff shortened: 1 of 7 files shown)


belisoful:
once this is in, I'll go through the Prado Modules and update them for dependencies.

belisoful:
I was considering a mini "helper" class to pass the id and require flag.  We could use a few of these micro-classes for data passing and storage.

ctrlaltca:
Sorry for the delay in merging these, but given it touches some delicate parts of the framework and the added complexity, i preferred to test it for a couple of days.

belisoful:
I figured.  And I appreciate it.

One of the reasons I did the TApplication refactor was to create more "order" within these delicate parts to make them easier to adjust.   It also gave me time to tweak the data structures to be more useful, even if parts are not used yet.

PR #1130, opened 2026-05-23
4.3.3 Final Working Memory and HISTORY.md

belisoful:
Final touches on the Doc Blocks (all configurable modules and services have xml and php examples).
Updated the Working Memory, and fixed some hyperlinking issues.
Updated the HISTORY.md.


belisoful:
The HISTORY.md looks "full" for 4.3.3. it's not clean and direct like prior entries.  I think it should be cleaned up, but I'm not sure how to do that.

belisoful:
I'll selectively commit these to master after this is (corrected and) merged.

Issue #1131, opened 2026-05-24
TReCaptcha2::getResponse clashes with TApplicationComponent::getResponse


belisoful:
https://github.com/pradosoft/prado/blob/7082e7f4a9ae0964df637630f796b16c3677a73e/framework/Web/UI/WebControls/TReCaptcha2.php#L279

TReCaptcha2::getResponse clashes with TApplicationComponent::getResponse.  It will not survive typing.

Changing to TReCaptcha2::getCaptchaResponse.

PR #1132, opened 2026-05-24
TReCaptcha2::getResponse to getCaptchaResponse - fix for #1131 - unit tests

belisoful:
`TReCaptcha2::getResponse` clobbers `TApplicationComponent::getResponse`.  They have incompatible return types.

`TReCaptcha2::getResponse` changed to `TReCaptcha2::getCaptchaResponse` for #1131


belisoful:
when the Prado::using bug gets fixed, this unit test will pass.

belisoful:
and another end of month bug in the unit test.

This fixes the run error in other TPropertyValue PR

Issue #1098, opened 2026-05-09
Transition of Javascript from "old-style" to ES6 (less dependent on jQuery)


belisoful:
One of the reasons I wanted to have the Prado Javascripts unit tests is for the possible transition to ES6. 
jQuery begins to take a less important roll.  We needed JS unit tests regardless and it was fruitful.

jQuery may throw off developers from other node.js environments.  jQuery is not ES6 compliant but we may be able to integrate the two.
There are some standard HTML 5 tags that replace some of the jQuery implementations.

jQuery is nice, it is also ancient.  Replacements for 2026 would be nice to noodle on.

belisoful:
We certainly shouldn't upgrade to ES6/ESM in a point release.  This is a good potential for v4.4.
the JS vi unit tests should be great for assisting this conversion and future JS developments.

belisoful:
There are more than a few changes that I'd like to "save" for 4.4.   New classes. new features, etc.   I think the changes to Data modernization should be in 4.4 as that comes along.

belisoful:
Thank you, @ctrlaltca

PR #1154, opened 2026-05-28
Prado JS Conversion from ES5 to ES6

belisoful:
This replaced jQuery with standard Javascript. Javascript has come a long way since jQuery started. jQuery was to help with the weak points of JS, except those holes are filled in over the last two decades.

There are still some places where jQuery is needed but much less so with this PR.

The conversion of jQuery to something more modern is the reason for the JS eslint, vitest, and playwright introduction.  And it was instrumental in terms of "the goal" to pass all the tests in the conversion.

#1098

### DIFF
--- a/framework/Web/Javascripts/source/prado/activecontrols/activecontrols3.js
+++ b/framework/Web/Javascripts/source/prado/activecontrols/activecontrols3.js
@@ -129,9 +129,9 @@ Prado.WebUI.TJuiAutoComplete = Prado.Class(Prado.WebUI.TActiveTextBox,
 	getUpdatedChoices(request, callback) {
         const lastTerm = this.extractLastTerm(request.term);
 		const params = new Array(lastTerm, "__TJuiAutoComplete_onSuggest__");
-		const options = Object.assign(this.options, {
-			'autocompleteCallback' : callback
-		});
+		// Stash the autocomplete-widget callback on this.options so onComplete
+		// can invoke it after the server round-trip resolves.
+		Object.assign(this.options, { autocompleteCallback: callback });
 		Prado.Callback(this.options.EventTarget, params, this.onComplete.bind(this), this.options);
 	},
 
--- a/framework/Web/Javascripts/source/prado/activecontrols/activecontrols3.js
+++ b/framework/Web/Javascripts/source/prado/activecontrols/activecontrols3.js
@@ -3,7 +3,7 @@
 /**
  * Generic postback control.
  */
-Prado.WebUI.CallbackControl = jQuery.klass(Prado.WebUI.PostBackControl,
+Prado.WebUI.CallbackControl = Prado.Class(Prado.WebUI.PostBackControl,
 {
 	onPostBack(options, event) {
 		const request = new Prado.CallbackRequest(options.EventTarget, options);
@@ -15,13 +15,13 @@ Prado.WebUI.CallbackControl = jQuery.klass(Prado.WebUI.PostBackControl,
 /**
  * TActiveButton control.
  */
-Prado.WebUI.TActiveButton = jQuery.klass(Prado.WebUI.CallbackControl);
+Prado.WebUI.TActiveButton = Prado.Class(Prado.WebUI.CallbackControl);
 /**
  * TActiveLinkButton control.
  */
-Prado.WebUI.TActiveLinkButton = jQuery.klass(Prado.WebUI.CallbackControl);
+Prado.WebUI.TActiveLinkButton = Prado.Class(Prado.WebUI.CallbackControl);
 
-Prado.WebUI.TActiveImageButton = jQuery.klass(Prado.WebUI.TImageButton,
+Prado.WebUI.TActiveImageButton = Prado.Class(Prado.WebUI.TImageButton,
 {
 	onPostBack(options, event) {
 		this.addXYInput(options, event);
@@ -34,7 +34,7 @@ Prado.WebUI.TActiveImageButton = jQuery.klass(Prado.WebUI.TImageButton,
 /**
  * Active check box.
  */
-Prado.WebUI.TActiveCheckBox = jQuery.klass(Prado.WebUI.CallbackControl,
+Prado.WebUI.TActiveCheckBox = Prado.Class(Prado.WebUI.CallbackControl,
 {
 	onPostBack(options, event) {
 		const request = new Prado.CallbackRequest(options.EventTarget, options);
@@ -46,19 +46,19 @@ Prado.WebUI.TActiveCheckBox = jQuery.klass(Prado.WebUI.CallbackControl,
 /**
  * TActiveRadioButton control.
  */
-Prado.WebUI.TActiveRadioButton = jQuery.klass(Prado.WebUI.TActiveCheckBox);
+Prado.WebUI.TActiveRadioButton = Prado.Class(Prado.WebUI.TActiveCheckBox);
 
 
-Prado.WebUI.TActiveCheckBoxList = jQuery.klass(Prado.WebUI.Control,
+Prado.WebUI.TActiveCheckBoxList = Prado.Class(Prado.WebUI.Control,
 {
 	onInit(options) {
 		for(let i = 0; i<options.ItemCount; i++)
 		{
-			const checkBoxOptions = jQuery.extend({}, options,
-			{
+			const checkBoxOptions = {
+				...options,
 				ID : `${options.ID}_c${i}`,
 				EventTarget : `${options.ListName}$c${i}`
-			});
+			};
 			new Prado.WebUI.TActiveCheckBox(checkBoxOptions);
 		}
 	}
@@ -69,14 +69,14 @@ Prado.WebUI.TActiveRadioButtonList = Prado.WebUI.TActiveCheckBoxList;
 /**
  * TActiveTextBox control, handles onchange event.
  */
-Prado.WebUI.TActiveTextBox = jQuery.klass(Prado.WebUI.TTextBox,
+Prado.WebUI.TActiveTextBox = Prado.Class(Prado.WebUI.TTextBox,
 {
 	onInit(options) {
 		this.options=options;
 		if(options['TextMode'] != 'MultiLine')
 			this.observe(this.element, "keydown", this.handleReturnKey.bind(this));
 		if(this.options['AutoPostBack']==true)
-			this.observe(this.element, "change", jQuery.proxy(this.doCallback,this,options));
+			this.observe(this.element, "change", this.doCallback.bind(this, options));
 	},
 
 	doCallback(options, event) {
@@ -90,13 +90,13 @@ Prado.WebUI.TActiveTextBox = jQuery.klass(Prado.WebUI.TTextBox,
  * TJuiAutoComplete control.
  */
 
-Prado.WebUI.TJuiAutoComplete = jQuery.klass(Prado.WebUI.TActiveTextBox,
+Prado.WebUI.TJuiAutoComplete = Prado.Class(Prado.WebUI.TActiveTextBox,
 {
 	initialize(options) {
 		this.options = options;
 		this.observers = new Array();
 		this.hasResults = false;
-		jQuery.extend(this.options, {
+		Object.assign(this.options, {
 			source: this.getUpdatedChoices.bind(this),
 			select: this.selectEntry.bind(this),
 			focus() {
@@ -129,7 +129,7 @@ Prado.WebUI.TJuiAutoComplete = jQuery.klass(Prado.WebUI.TActiveTextBox,
 	getUpdatedChoices(request, callback) {
         const lastTerm = this.extractLastTerm(request.term);
 		const params = new Array(lastTerm, "__TJuiAutoComplete_onSuggest__");
-		const options = jQuery.extend(this.options, {
+		const options = Object.assign(this.options, {
 			'autocompleteCallback' : callback
 		});
 		Prado.Callback(this.options.EventTarget, params, this.onComplete.bind(this), this.options);
@@ -159,17 +159,25 @@ Prado.WebUI.TJuiAutoComplete = jQuery.klass(Prado.WebUI.TActiveTextBox,
 
 
 	onComplete(request, result) {
-  		const that = this;
-  		if(that.options.textCssClass===undefined)
-  		{
-			jQuery.each(result, (idx, item) => {
-				result[idx]['value']=jQuery.trim(jQuery('<div/>').html(item['label']).text());
-			});
-  		} else {
-			jQuery.each(result, (idx, item) => {
-				result[idx]['value']=jQuery.trim(jQuery('<div/>').html(item['label']).find(`.${that.options.textCssClass}`).text());
-			});
-  		}
+		// Decode the HTML in `label` to plain text for the underlying value.
+		// Uses a detached <div> as the HTML parser; matches the old
+		// jQuery('<div/>').html(label).text() pattern.
+		const decode = (html) => {
+			const div = document.createElement('div');
+			div.innerHTML = html;
+			return div;
+		};
+		if(this.options.textCssClass === undefined) {
+			for (const item of result) {
+				item.value = decode(item.label).textContent.trim();
+			}
+		} else {
+			const sel = `.${this.options.textCssClass}`;
+			for (const item of result) {
+				const node = decode(item.label).querySelector(sel);
+				item.value = (node ? node.textContent : '').trim();
+			}
+		}
 
 		request.options.autocompleteCallback(result);
 	}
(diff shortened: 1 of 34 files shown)


framework/Web/Javascripts/source/prado/activecontrols/activecontrols3.js
```diff
@@ -3,11 +3,10 @@
 /**
  * Generic postback control.
  */
-Prado.WebUI.CallbackControl = jQuery.klass(Prado.WebUI.PostBackControl,
+Prado.WebUI.CallbackControl = Prado.Class(Prado.WebUI.PostBackControl,
```
belisoful, line 6: New Prado Abstraction

framework/Web/Javascripts/source/prado/activecontrols/activecontrols3.js
```diff
@@ -3,11 +3,10 @@
...
 {
-	onPostBack : function(options, event)
-	{
-		var request = new Prado.CallbackRequest(options.EventTarget, options);
+	onPostBack(options, event) {
```
belisoful, line 8: There is a lot of this kind of conversion

belisoful:
I've been over this code.  It makes sense.  We need to move to the new JS standards.

The js unit tests pass. the functional tests all pass.  The only thing left is 3rd party human verification.

This may be the start of the end of jQuery in Prado... maybe...   We still have some pretty strong dependencies in places.  I'm not suggesting it be removed, I'm just pointing out that it's not as important these days.

framework/Web/Javascripts/source/prado/controls/htmlarea.js
```diff
@@ -59,29 +56,25 @@ Prado.WebUI.THtmlArea = jQuery.klass(Prado.WebUI.Control,
...
 
-	compressedScriptsLoaded: function()
-	{
+	compressedScriptsLoaded() {
 		Prado.WebUI.THtmlArea.tinyMCELoadState = 255;
-		var wrapper;
+		let wrapper;
 		while(Prado.WebUI.THtmlArea.pendingRegistrations.length>0)
-			if (wrapper = Prado.Registry[Prado.WebUI.THtmlArea.pendingRegistrations.pop()])
+			if ((wrapper = Prado.Registry[Prado.WebUI.THtmlArea.pendingRegistrations.pop()]))
```
ctrlaltca, line 63: Does this make any difference?

ctrlaltca (approved):
Wow, this is huge, it was on my TODO-list since years.
I checked out the code and it looks ok.
I guess we can remove jQuery as a dependency on a few packages here:
https://github.com/pradosoft/prado/blob/master/framework/Web/Javascripts/packages.php#L111
I'll give it a run on a few big webapps to see if any convenience compatiblity layer can be added to help keeping custom controls running without changes.

belisoful:
I was considering removing jQuery as a dependency in a few packages.

The question is: _if developers should be expected to add it if they depend on it - depending on the packages depending on it._  I was leaving this to the community and discussion before doing that.

It sounds like you are in favor of removing the dependencies.  I agree, but wanted to count hands.

@ctrlaltca **What else is on your TODO list?**

This was the exact conversion I was working towards with node.js, eslint, vitest unit tests, and playwright. It was why I wanted everything covered.  It needed to have a really good goal to reach for. 

I started this conversion 3 days ago.  It took about 2 hours of Claude tokens in sessions over that time. First I asked for a plan, when it was tuned, I  gave it my approval to covert in 5 phases.  The initial work took about an hour with it cycling until all unit and functional tests pass.  It commented that the unit tests were instrumental in making the transition quick and smooth.

The last month has been an example of $100 sub @ claude.ai.  Not an ad.  I'm just saying.  Trust it and the process of coding is more like tuning and shaping.  The limiting factor becomes how quickly can we read and process the code on different scales.  The efficiency it provides is unreal, as you can see.

belisoful:
Take your time proofing this. There is no rush.  The Active versions of the HTML5 controls are far from here.

BTW, doing a cheeky `return $this` on the tail end of every property setter is something I'd like to see happen sooner rather than later.

ctrlaltca:
From my few tests done today all builtin js controls seems to work just fine.
My custom controls based on jQuery are mostly working due to the jQuery.klass compatibility layer to Prado.Class , but they are exposing https://github.com/jquery/jquery/issues/2476 . I'll try to explain the problem.
 
In Prado js we use `Prado.Class.observe` to add an event handler.
Before this PR events this method was using jQuery's event handling:
https://github.com/pradosoft/prado/pull/1154/changes#diff-10517f069d0d64bf6c10f5cb96fb9e1933788b48e0b136d27c1d348515532aeaL87
```js
return jQuery(e._element).bind(e._eventName, options, e._handler);
```
while now it's using standard js event handling:
```js
target.addEventListener(eventName, handler);
```

Unfortunately we didn't implement a `Prado.Class.trigger` method to trigger a registered event, so my custom controls (as old prado controls) were using `jQuery.trigger` :
```js
jQuery(this.control).trigger("change");
```
Unfortunately this function is not able to fire js event listeners, only jQuery events. So these calls just become no-ops.
The obvious solution is just to replace these calls with
```js
this.control.dispatchEvent(new Event('change', { bubbles: true }));
```
but this means rewriting all custom js controls.
I'd like to keep it backwards compatibile as much as i can, so here's a possible solution:
```js
(function ($) {
  const originalTrigger = $.fn.trigger;

  $.fn.trigger = function (type, extraParameters) {
    const isEventObject = type && typeof type === 'object';
    const eventType = isEventObject ? type.type : type;

    return this.each(function () {
      const el = this;

      const klassObject =
        Prado.Registry[el.id] && Prado.Registry[el.id].constructor.name == 'klass';

      if (!klassObject || !el.dispatchEvent || !eventType) {
        originalTrigger.call($(el), type, extraParameters);
        return;
      }

      let nativeEvent;

      if (isEventObject && type instanceof Event) {
        nativeEvent = type;
      } else {
        nativeEvent = new Event(eventType, {
          bubbles: true,
          cancelable: true,
          detail: extraParameters
        });
      }

      el.dispatchEvent(nativeEvent);
    });
  };
})(jQuery);
```

This method overrides jquery.trigger by checking if the calling object is a "klass" and triggering a native event.
Non-prado standard query controls are not "klass" so they keeps using jquery event handling.
Do you think it's reasonable to add such a workaround?

ctrlaltca:
> @ctrlaltca **What else is on your TODO list?**

I don't really have a written list, but here's a few things coming up at my mind: 
 * Creating new html5 based controls to replace eg. TDatePicker, TColorPicker, etc... with 1:1 method signatures, so you can just plug them in replacing the old name with the news one and benefit from them
 * move all the database classes and handling in a separate library (so people can plug in their favorite ORM; eg Doctrine, Propel, ...)
 * i the past i thought about adhering to https://www.php-fig.org/psr/ interfaces and factories (eg the cache interface), but given what happened to Yii during the Yii2 to Yii3 migration i abandoned this idea.

belisoful:
> This method overrides jquery.trigger by checking if the calling object is a "klass" and triggering a native event.
> Non-prado standard query controls are not "klass" so they keeps using jquery event handling.
> Do you think it's reasonable to add such a workaround?

it is reasonable.  I have some company for a few days.  My output won't be like it's been this month for a few days.

I'm working on the HTML5 stuff too.  When we went to 4.4-dev, I put the html5 stuff aside for a few more core changes, like coercion, dependencies, and PradoUnit.  I'm with you on the conversion of WebControls to the modern standard.

and what happened with Yii2 to Yii3 with PSR that was such an issue that you've reconsidered?

PS. 4.3.3 is only pending minor edits and new Interfaces for separate libraries like Doctrine and Propel. I'm testing it with my MongoDB extension to ensure that it works.  4.3.3 may have the interfaces, but 4.4 will be refactored to fully integrate other ORM.

ctrlaltca:
> and what happened with Yii2 to Yii3 with PSR that was such an issue that you've reconsidered?

Nothing bad, yii3 libs are awesome pieces of code and samdark made a hell of a work on them.
Still, the framework had to be split in dozens of different subprojects and it took a very long time (6 years) for it to be actually released.
I worked with both yii2 and yii3 "alpha" in the past. Do you imagine i even had an alpha version of prado 4 rebased on yii libraries, able to run some basic apps like prado-demos without needing any code change?
Still, the compatibility layer was so huge that i dropped the idea after working on it for some months.

belisoful:
I don't see how Prado4 could be rebased to yii3.

Is that a direction you envision?   There are a few things Prado does better.   TEventHandler is a prime example.  Yii is  trying too hard in a few ways.  They've got some things right too.  What would you re-write to be rebased in Yii3?

I'm attempting to make the 4.3.3 Data capable of having Plugins for other ORM, as you say, but also within the goal of Backward Compatibility (BC).   We won't make in all the way in 4.3.3, but that's my goal for Data v4.4.

It is in the process of writing Mongo integration tests for Prado ActiveRecord (AR).   The goal is MongoDB as a drop in for AR, DataGateway, common, and basic Data classes.    If Mongo can be a drop in, then it should work for the rest.

BTW, AR data source of truth was the Db Columns.   Mongo doesn't have columns- it's categorized as NoSql.  MongoDB will not work with SqlMap (MyBATIS) for obvious reasons.  The change in AR for this is to use the PHP object properties as the source of truth for databases flagged in an interfaces as such.

NoSql- everything is metadata, all AR object properties are loaded and saved.  NoSql is really convenient for Prado AR.

Regarding the JS event system.  Let me audit the code for the cross compatibility of JS vs jQuery.

...  If you could "rewriting all custom js controls" and maintain BC, would you choose that option?

as you see, the agents I'm using are pretty powerful with a knowledgeable user.

ctrlaltca:
> I don't see how Prado4 could be rebased to yii3.
> Is that a direction you envision?

Nope, it was just an experiment. At that time prado 3.x accumulated a huge technical debt, so i could either try to update it (with the adoption of composer, jquery, etc..) or create a "migration layer" to yii for my existing apps.
I tried both paths, choose the first one and discarded the failed experiment.

> There are a few things Prado does better. TEventHandler is a prime example. Yii is trying too hard in a few ways. They've got some things right too. What would you re-write to be rebased in Yii3?

The idea of replacing some prado code with external libraries (from yii or even symfony) was an idea to lower the maintenance burden. I've been quite alone in maintaining prado for a long time, working on it during the night.
Nowadays the situation is really different, i can barely keep up with your work :)

> I'm attempting to make the 4.3.3 Data capable of having Plugins for other ORM, as you say, but also within the goal of Backward Compatibility (BC). We won't make in all the way in 4.3.3, but that's my goal for Data v4.4.

I would take the whole framework/Data folder (and the related classes, eg TDbCache) to an external library, eg "Prado ORM". Not all foreign ORMs adapts really well to the ActiveRecord/Scaffolding patterns.

> BTW, AR data source of truth was the Db Columns. Mongo doesn't have columns- it's categorized as NoSql. MongoDB will not work with SqlMap (MyBATIS) for obvious reasons. The change in AR for this is to use the PHP object properties as the source of truth for databases flagged in an interfaces as such.

That's some huge work I would never pick up.
To be transparent, i didn't use Prado's Data classes in any of my bigger projects, but only for some minor one.

> Regarding the JS event system. Let me audit the code for the cross compatibility of JS vs jQuery.
> ... If you could "rewriting all custom js controls" and maintain BC, would you choose that option?

Yes, definitely. Even maintaining only "some" BC.

belisoful:
@ctrlaltca Thank you for sharing!  Your story is amazing and compelling.  I know you've been a sole developer.  I've stepped in from time to time to make up technical debts, like the Cron Manager, Http Headers Manager, etc.

The stunning productivity has been directly from the $100/mo plan on Claude.  It's even shocked me. The issue has always been the speed of coding rather than my lack of imagination and where I envision Prado 4.4+/5.x. This is not prophetic, it is already starting to happen: Those who don't use these LLM tools are being left behind.

You can see my productivity.  Even the "weak" LLMs are tremendous.  Changing a constant name may affect 10-50 sites in the class, other code, doc blocks, and unit tests.  When it is told to change the variable name, it literally goes through every element for alignment.  That use to be so tedious, add lots of time to coding, and reason why the documentation was poor. but the LLM can make quick work of the whole thing in 1-5 minutes (based on complexity).  So even something so simple as changing the name of a variable, it's literally easier to have it address the change.  if I change the const name by hand, I have to tell it to audit the code, which usually takes longer to "fix/align" all the other sites where it is used.

I'm a bit surprised that you haven't started trying LLMs as a tool, @ctrlaltca.

I am tremendously lucky and glad that you are reviewing Prado PRs.  I know it is a lot, but the technical debt is real.  There a lot of catching up to do.  I'm doing my best to stay focus and isolate the changes. There is order, some PRs need to be in place before others. My limiting factor is reading and comprehension of the code these tools generate.  IT-LLMs are coding, devs are "shaping" the code through it.  I know what I want, or at least what a good end state looks like when we get there.

JUI and Active Controls do need a serious update- for 2026+.  The Control "Adaptor" subsystem basically could/should all be replaced by the behavior system, that's my basic off-the-cuff analysis.  But I don't want to step on your feet for all the amazing hard work you did with it.  The behavior system opens up a lot of pathways for implementation that were not in Prado when the ActiveControls and Jui were made.

Here is my bucket list:
- Get Prado Data working with 3rd party DB. (aligned with your vision)
- Upgrade Prado Data for insertOrIgnore and Upsert.
-  Upgrade Prado Data to be able to Create and alter tables (for TDb* classes that Auto create tables).
-  I'd like to split out Active Record from Prado into its own prado repo. (and upgrade to latest standards)
-  I'd like to split out SqlMap from Prado into its own prado repo. (and upgrade to latest MyBATIS standard)
- Ensure separation between Prado `I/TDb*` as Prado's PDO layer and Prado `I/TData*` as the Db-agnostic side.
- Upgrade WebControls to HTML, add missing HTML5 tags, upgrade the existing controls from custom to HTML5 where there is an HTML5 replacement (aligned with your vision)
- Upgrade IO, lots of new additions here.  potentially many many.
- Upgrade Assets Manager (aka make it the TPublishingManager), basically v2 of TAssetManager.  Already done, waiting for Basic IO.
- Multi-App mode (may require Proxy classes for global singletons like THttpRequest)
- more Testing Harness.
- Make THttpHeaderManager the single central point for http headers
- HTML 5 controls — new and update existing controls (drop in) (your vision as well)
- start working on PSR in Prado itself, in ways that make sense.
- add `behavior` to the app configuration rather than simply in via TBehaviorsManager.
- Upgrade TUserManager.
- Role Manager
- (lower priority) Upgrade Shell
- put types on all the getters and setters in the 48+ configurable modules and services in Prado
- put `return $this;` on all property setters (that which make sense) for chaining.
- Upgrade Cron to be user oriented rather than more admin oriented.  eg. a user in the website can register a future task.
- PHPAttributes for behaviors to register which of its methods are "for the owner", speed boost.
- Proxy classes, eg TCacheProxy (where primary is true) and TDataSourceConfigProxy,  the app modules can point at the proxy and the proxy forwards to the backing.  The backing can then be edited in one place rather than multiple
- JS upgrade
- upgrade all the unit tests to the Harness and Prado Unit.
- Apply Uniform Access Principal - Self Encapsulation (UAP-SE) to everything. This is where the "new" TApplication came from and it has helped tremendously with TShellApplication and TTestApplication.
- Abstract the Template system, internally for more pluggability.  eg.  maybe allow for 3rd parties to add their own render tags (like `<%- ... %>` or `<%+ ... %>` ).
- Abstract TPageService to have different "render systems", one of which is Prado's TControl/TPage.  It could be jacked for React rendering.
- HTTP header Authentication. (changes to TAuthManager also)
- Authentication Token Store Manager (needed for HTTP Header Auth)
- Integrity Manager (with additional file config)
- TStyleSheetAsset
- maybe an object transfer class (class name, properties, config, behaviors)
- new `TObject` base class without behaviors.
- Macroable on top of Behaviors?
- convert all the configurable modules/services to normalize their xml into php before parsing in php, one parse path.

Already merged:
- TFileCache and TMemoryCache (sorely needed)
- TModule::dyPostInit
- Promote "default" TApplication modules into module list.
- IModuleDependency - resolves `init` dependencies between modules
- THttpHeadersManager (and CSP with Reporting service). 
- Property Type Coercion
- playwright conversion
- JS unit tests (vitest, eslint)
- PradoUnit unit testing helper library.  (btw, this is **Massive** in terms of enabling of unit testing.  If you haven't seen it, I recommend you do)
- Unit Testing Harness
- Functional Testing Harness
- TTemplate and TStyle upgrades
- beginning of HTML5 controls
- starting of 3rd party Data.

ctrlaltca:
> I'm a bit surprised that you haven't started trying LLMs as a tool, @ctrlaltca.

Oh yes I did try them, eg. the code in https://github.com/pradosoft/prado/pull/1154#issuecomment-4583629942 was generated by an LLM and then edited to suit my needs.
Nowadays my work doesn't  involve coding a lot like in the past, that's also why prado development by me slowed down a lot in the last period.
Still , I'm using LLMs to assist my hobbies, mostly small electronics projects (esp32, nrf52) and reverse engineering tasks (get root on that new Chinese electronic thing, dump its firmware, etc..).

> Here is my bucket list:

Wow. This is amazing list of updates. 

> The Control "Adaptor" subsystem basically could/should all be replaced by the behavior system, that's my basic off-the-cuff analysis. But I don't want to step on your feet for all the amazing hard work you did with it.

Adaptors are a relic of the past, when everything had to fallback nicely on a non-js enabled browser. We're talking about times where IE was the predominant browser. I've tried numerous approaches to get them down the bin without breaking too much, but never found a satisfactory solution. I'm quite sure you will.

belisoful:
I understand.  Prado is both a hobby, a revival project, and something I'd like to more seriously use if not for the tech deficit.

Glad to know that Control Adapters are on the chopping block.

There are going to be a lot of changes.  Dare I say that we may even be preparing for Prado 5?  I don't think it'll take too long at this rate.

belisoful:
#1154 Trigger Fix - added cross dispatch and validation against double dispatch.
jQuery.fn.trigger BC update  - slightly more correct than the solution provided above
klass.prototype.trigger added - added in Prado Klass trigger.

PR #1157, opened 2026-05-29
WIP - Proxy classes - IProxy, TComponentProxy, TModuleProxy, TCacheProxy, TDataSourceConfigProxy

belisoful:
This is a very interest set of classes.

I'm not the biggest fan of TModuleProxy, but Cache and Data Source Config Proxy would be extremely useful.

The proxy can be the primary and Backing can be changed on the fly for all modules linked to the Proxy rather than changing each one separately.

Also, a subclass of TModuleProxy could serve as a TLazyModule that gives out the lazy proxy, and when it gets called, it would load the lazy module and start calling it through the proxy.

Some interesting options.


belisoful:
These may be needed for Multi-application mode regarding truly global singletons like THttpRequest.

PR #1167, opened 2026-05-31
TComposer - static helper class for composer dependency metadata

belisoful:
and a bunch of helpers and a trait that will help.


belisoful:
For the community:
BTW, this is a very interesting introspection class around Composer that we will be using in the future.

PR #1169, opened 2026-06-01
TPropertyValue::ensureArrayOfType and ensureNullIf

belisoful:
Adds:
- ensureNullIf- is ensureEmpty but split into each bit: `null`, `false`, `''` and all together
- ensureArrayOfType- coerces the array elements into a type with filter options, " above and filters for trim and lowercase on strings.
- FILTER_ options for 

Proper examples to the doc blocks.

Changes `self::` to `static::` in proper places.


belisoful:
This is a pretty big unlock btw.

PR #1185, opened 2026-06-04
4.4.0 composer adds php-cs-fixer and phpstan

belisoful:
and updates the weak functional tests

### DIFF
--- a/framework/Web/TAssetManager.php
+++ b/framework/Web/TAssetManager.php
@@ -143,6 +143,40 @@ use Prado\Util\Traits\TInitializedTrait;
  * ]);
  * ```
  *
+ * Pattern matching follows gitignore-style rules. A pattern without a slash matches
+ * the file name at any depth (`*.js` matches `js/app.js`). A pattern containing a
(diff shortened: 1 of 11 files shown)


framework/Web/TAssetManager.php
```diff
@@ -207,21 +207,11 @@ public function init($config)
 		if ($this->_baseUrl === null) {
 			$this->_baseUrl = rtrim(dirname($application->getRequest()->getApplicationUrl()), '/\\') . '/' . self::DEFAULT_BASEPATH;
 		}
-		$this->setAppAssetManager();
+		$application->setAssetManager($this);
```
ctrlaltca: This looks like a partial revert of #1179 
I'm not sure it was intentional

ctrlaltca (changes requested):

framework/Web/TAssetManager.php
belisoful: 🙏 There are lots of copies of the repo.

belisoful: fixed

PR #1193, opened 2026-06-06
TDatePicker - InputMode renamed to DateInputMode

belisoful:
due to collision with new TWebControl::InputMode

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -14,6 +14,7 @@ Upgrading from v4.3.3
 - Prado JS dispatches `prado:ajaxComplete` event with jQuery.onAjaxComplete for non-dependence on jQuery. Use this event instead of the jQuery event.
 - `Prado\Web\TUri` is now a PSR-7 `UriInterface`: scheme/host are lower-cased, path/query/fragment are
   percent-encoded, and `getPort()` returns `?int` (`null` when absent or scheme default).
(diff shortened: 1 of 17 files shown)


ctrlaltca:
I don't think it's a good idea breaking all the existing apps using a T*DatePicker.
Can you provide more context on the new TWebControl::InputMode you want to implement?

belisoful:
HTML 5 has a global tag directly called "inputmode" and TDatePicker collides with it.  TWebControl is getting HTML5 capabilities, but TDatePicker is basically a blocker.  We can wait on this PR until you see the TWebControl update for HTML5.  They do kind of go together, while unrelated.  I wanted to isolate this PR from the TWebControl HTML5 update, which is big enough on its own.

Here is the reference and the context for adding InputMode to TWebControl:
[HTML5 global attribute - inputMode](https://www.w3schools.com/tags/att_global_inputmode.asp)

The only other option is to continue allowing TDatePicker to override InputMode; TWebControl::InputMode is not accessible for TDatePicker. 

The `Data` attribute for html 5 global attributes conflicts with critical TData* controls but can be resolve as "Dataset" as that is the javascript name for the same.  But InputMode as a global html attribute cannot be tweaked like Data=>Dataset.

ctrlaltca:
Thank you for clarifying the issue.

PR #1195, opened 2026-06-07
TWebControl HTML5 Update - TWebAttributeCollection, TWebInputMode,

belisoful:
TWebAttributeCollection - tracks `data-` (`TWebControl::Dataset`) and `aria-` (`TWebControl::Aria`) attributes, accessible without prefix.
TEnterKeyHint
TWebControl - adds:
- Dir
- Hidden
- Lang
- Spellcheck
- draggable
- inert
- ContentEditable
- EnterKeyHint
- Translate
- Popover
- Role
- ARIA Attributes
- Data Attributes (named `Dataset`, after the JS version) -  Note: Data collides with TData* WebControls


ctrlaltca:
Hey, I was already reviewing this out of curiosity before you set it as ready :)
Looks good :+1:

belisoful:
It's been sitting in my queue of work since the TStyle stuff was submitted.   There have been other more significant things than this that were needed for v4.4 before this.  like the TPropertyValue work.  There are still some pretty cool things on the way.

PR #1207, opened 2026-06-09
TDot - cs-fixer tweak

belisoful:


ctrlaltca:
php-cs-fixer is always picky on these things :)

PR #1208, opened 2026-06-10
.gitignore organizing + remove composer.lock per .gitignore

belisoful:
.gitignore was organized into sections and old elements removed.

NOTE: composer.lock was accidentally checked in, a pain to maintain, and is in .gitignore.

    **`composer update` needs to be called after this**

It only shows up as different (despite being in `.gitignore`) when it's changed because it was already there.  It will be ignored once it is removed.


belisoful:
Don't forget to `composer update` or `composer devtools` after pulling this commit.

PR #1215, opened 2026-06-13
RPC Service and RPC Client moved to repo `pradosoft/prado-rpc`

belisoful:
XML-RPC is deprecated and not included as an extension in PHP 8.

It does require its own xml-rpc library for effective plug and play.

see https://github.com/pradosoft/prado-rpc


belisoful:
This is waiting on proper config for extensions.  Configuration of extensions for errorMessages and classMaps happens when the extension is pulled in via TApplicationConfiguration (by composer reflection), but code execution is opt-in by application configuration file.

This makes a `require $phpFile` a security issue for extensions for their class map.   I'm working on a json class map for composer extensions.  This way it gets loaded like php, but is data/parsed - like xml.

belisoful:
@ctrlaltca https://github.com/pradosoft/prado-rpc is basically ready.  It needs to be tagged and released then added to packagist.org.  It should be reviewed before tagging.

When it is done, this PR will be marked "Ready" and the RPC classes removed.

PR #1217, opened 2026-06-13
TSocketReactor - socket event loop

belisoful:
TSocketReactor - The manager of `stream_select` IO, independent 
TSocketServer::bind - instances static rather than self
TSocketStream - adds getCryptoMeta and getAlpnProtocol

### DIFF
--- a/framework/IO/Socket/TSocketAddress.php
+++ b/framework/IO/Socket/TSocketAddress.php
@@ -39,6 +39,10 @@ class TSocketAddress extends TComponent
 	/** @var ?string The path (unix/udg sockets), or null. */
 	private ?string $_path;
 
+	// =========================================================================
+	// Construction and Parsing
+	// =========================================================================
+
 	/**
 	 * Builds a socket address from its components.
 	 * @param ?string $scheme The transport scheme.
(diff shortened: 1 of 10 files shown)


ctrlaltca:
Wow, this reproducibly crashes php 8.1!

belisoful:
I hadn't checked back to see the results...  been busy with other stuff.   Let' me check it out...

belisoful:
TSocketServer::addConnection() attaches onClose/onFinalize handlers (static closures holding a \WeakReference to the server) to every tracked connection. When a connection is still in the registry as the server is destroyed, the engine frees _connections during property teardown, which frees that connection, whose destructor fires onClose, whose closure resolves the weakref back to the half-destroyed server and calls removeConnectionDirect() — unset($this->_connections[...]) on the very array the engine is mid-freeing. That's a use-after-free → heap corruption, surfacing later as a crash in emalloc/the malloc free list.

PHP 8.1's WeakReference::get() still resolves to an object during its own destruction; 8.2+ returns null, so the closure no-ops there. 8.1 only.

ctrlaltca:
Wow, thank you for the fix and the explanation

belisoful:
@ctrlaltca This merge enabled this repo: https://github.com/belisoful/prado-websockets

a web connection can get upgraded to a TWebSocketService, and it has a TWebSocketServer for browsers to connect to.  as in - standalone background socket server on its own process.

PR #1222, opened 2026-06-14
TTime - <time/> tag

belisoful:
TTime — HTML5 <time> element control

Adds TTime and TTimeFormat to render the HTML5 <time> element with a machine-readable datetime attribute and human-readable visible text.

TTimeFormat — 20-constant TEnumerable covering every WHATWG valid datetime value category: time strings, date strings, month, week, yearless date, year, local datetime (space- and T-separated), and global datetime with timezone offset.

TTime — TI18NWebControl that accepts DateTime as a DateTimeInterface, DateInterval, numeric Unix timestamp, or parseable string. The DateTimeFormat property selects the datetime="" attribute encoding. Visible text resolves in priority order:

- TextFormat property — a TTimeFormat name or ICU datetime pattern.
- Child content interpreted as a TTimeFormat name or ICU pattern; falls back to rendering children as-is.
- DateTimeFormat value formatted as human-readable text.

Negative DateInterval values are rendered with a "-" prefix in visible text (HTML5 has no negative duration form; the datetime attribute stays positive).

TIntlDateFormatterTrait — extended getIntlDateFormatter() with an optional $pattern parameter and a fourth cache level, enabling pattern-keyed formatter reuse across TTime, TSimpleDateFormatter, and TDateFormat.

Other: two error codes added to messages.txt; framework/classes.php updated; 120 unit tests in TTimeTest and TTimeFormatTest.

### DIFF
--- a/framework/Exceptions/messages/messages.txt
+++ b/framework/Exceptions/messages/messages.txt
@@ -670,6 +670,9 @@ dbcron_connectionid_invalid				= TDbCronManager.ConnectionID '{0}' is not a TDat
 
 timescheduler_invalid_string			= TTimeScheduler could not parse the schedule element '{0}'
 
+time_invalid_datetime					= TTime DateTime must be a DateTimeInterface, DateInterval, string, or numeric value; {0} given.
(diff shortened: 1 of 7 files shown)


belisoful:
The finicky test is resolved in #1223.

PR #1228, opened 2026-06-22
IO Socket Server - TSocketServerModule, TSocketServerAction

belisoful:
This is a prebuilt socket server module and cli action to trigger it.  Both need overriding to be fully functional.


ctrlaltca:
Hey @belisoful sorry for the delay, i'm on vacation with limited access to the interwebs

belisoful:
@ctrlaltca That's what I was hoping!  You deserve a break.  I'm ok with the delay. 🙂  I'm spending some time on other projects now too.   Some of the tech deficit in PRADO is made up, so I'm pulling it back just a little.  It's still going strong, just a bit slower.

My other projects: Prado extensions (http2, web socket, mongo db, image, etc), an Apple Metal (GPU) video/image filter and compute library, a Final Cut Pro FxPlug framework, shaderToy library for MacOS and iOS.

There should be some nice software packages that can be unlocked with these libraries when complete.

PR #1235, opened 2026-07-30
TDialog, TActiveDialog, TDetails, TSummary, TActiveDetails

belisoful:
Implements the `<dialog>` and `<details>` HTML5 tags.


belisoful:
Should I break this up into two PRs?  They are very similar in terms of the HTML5 tags which is why I put them together.

ctrlaltca:
> Should I break this up into two PRs? They are very similar in terms of the HTML5 tags which is why I put them together.

Don't worry, as long as it's new code and it doesn't touch existing code it's ok

PR #1239, opened 2026-08-13
IO/Compression/ - Standardized interface for compression and decompression of data

belisoful:
Implementing it lets a caller select a compression codec without binding to one algorithm.


belisoful:
There is a question here on what to do:
    I'd like to break out the `prado-io` into its own repo as it completes.  Should the Compressors be part of the IO repo or should they have their own separate repo, like `prado-compressors`?
    
If the compressors are part of io, then they should be included here for the move.
otherwise compressors should be their own repo.
(My preference is a ever so slightly towards separate repo)

Another question:  Should we have an XZ Compressor as a stub for a future possible PHP repo for it?  (My preference is yes, because of the next question)
And lastly, should the PRADO classes for the built in compressors also attempt to use the command line versions? (3rd party can supply cli compressor support, [I was considering filling this gap])

PRADO is the layer for and on top of PHP, not OS cli dependent.

ctrlaltca:
The "modern way" is to try to split everything up in a lot of small, more manageable repos and including them.
PRADO has always been the exact opposite of this, including everything in a single monolithic framework.
But i agree some parts of the framework cloud be split and moved to a separate repo, eg.:
 * really specific components, eg. some of the IO classes you are creating in the last period
 * the whole Data / PDO and scaffolding components
 * components requiring external dependencies
 * components we are phasing out but still support (eg. the old pradosoft/prado-tinymce3)

The final decision for the IO/compressor repo is up to you.
About using cli commands, i'm not really a fa of this approach, but if it's the only way to implement something i see no better option.

belisoful:
I am just a little torn between the old monolithic way and modern "split everything" way.  I want to go with the trends.   That may be a PRADO 5 thing, to break up prado into smaller chunks.

OK.  The Compressor classes go in IO.  I'm going to include an XZ stub for the future. They will only be based on PHP and PHP extensions.  This is a similar seam split as Prado Data, only supporting what is in PHP.  I'll do my own repo that supports the cli version.  It will fit well as a Fallback for the standard PHP compressors, though XZ is CLI only at this point.

belisoful:
@ctrlaltca [prado-demos repo](https://github.com/pradosoft/prado-demos/pull/8) has two bug fixes.  I am leaving in the composer.lock so it doesn't need to be re-created with each clone.  prado-demos#8 is for the up coming TWebTemplate control `<template>` tag.  It is the first new client side JS control in a long while.

belisoful:
@ctrlaltca One more on https://github.com/pradosoft/prado-demos/pull/10 to keep the TWebTemplate dev happy.

belisoful:
FYI, anyone interested: I have a Compression fallback library here that uses cli when the php compression library is missing.
https://github.com/belisoful/prado-compression

ganiuszka:
Hello @belisoful 

This is really good idea with this fallback library. In my opinion this should be even faster than PHP compression extensions. It's nice in case bigger files. Thanks!

PR #1240, opened 2026-08-19
IO/Stream/ - PSR-7 utility stream sources and decorators - unit tests, classes.php, messages.txt

belisoful:
Adds the utility stream family to `Prado\IO\Stream`: sources that adapt non-stream
producers to PSR-7, and decorators that reshape an inner stream, modeled on the
Guzzle PSR-7 utility set.

| Class | Role |
|-------|------|
| `TAppendStream` | Reads several streams in sequence as one; read-only. |
| `TMultipartStream` | Builds a `multipart/form-data` body as an append stream. |
| `TPumpStream` | Read-only stream produced on demand by a source callable. |
| `TFnStream` / `TFnStreamMethod` | Stream from a map of per-method closures, keyed by the enumerated PSR-7 method names. |
| `TLazyOpenStream` | Defers opening the inner stream until first use. |
| `TCachingStream` | Makes a non-seekable source seekable by caching reads; read-only. |
| `TLimitStream` | Presents a byte window (offset + length) of an inner stream. |
| `TDroppingStream` | Caps the inner stream's size, dropping excess writes. |
| `TNoSeekStream` | Hides the seekability of an inner stream (seek and rewind throw). |

Notes:
- PSR-7 strict: closed/detached streams are unusable (reads throw, `isReadable()`
  false, `getSize()` null), `__toString()` never throws, and unsupported
  seek/write/read operations throw rather than silently no-op.
- The callable-holding streams (`TFnStream`, `TPumpStream`, `TLazyOpenStream`)
  refuse unserialization, so a crafted serialized payload cannot choose what runs.
- `TMultipartStream` sanitizes names, filenames, and extra headers against header
  injection (CR/LF and quote escapes).
- All classes registered in classes.php; `multipartstream_element_invalid` added
  to messages.txt.
  
  Final classes of #904.


belisoful:
This is the remaining classes for Prado IO.  There is still some remaining clean up and conformance to perfect, but the whole Prado IO is basically now in place with this PR.

PR #1243, opened 2026-08-26
TDbPropertiesTrait - null guards on $app for module and runtime

belisoful:
This tweaks the TDbPropertiesTrait to ensure that db properties will work if there is no app (eg for unit testing)


belisoful:
If there is anything needed to clear this, let me know.

PR #1251, opened 2026-08-28
TSafetyCover - a panel with a protective cover over an area  to prevent stray clicks

belisoful:
This is an unusual class.  I don't think there is an equivalent "control" in other frameworks.

# TSafetyCover — click-to-confirm cover control

Adds `TSafetyCover` (since 4.4.0).
A `TPanel` whose body content sits behind a click-to-open cover, guarding controls
that act immediately (e.g. a delete button) against accidental activation.

https://github.com/user-attachments/assets/5042ab1b-e94d-4a6b-a5ef-b21ba8f0f078

## Features

- **Two-layer guard** — a transparent `pointer-events` guard re-blocks the content
  the instant it closes while a visible *face* animates; no click-through during
  the close.
- **Open effects** — `OverlayEffect` = `Slide` / `Collapse` / `None`, any
  `OverlayDirection` (`Up`/`Down`/`Left`/`Right` + RTL-aware `Forward`/`Backward`),
  independent `OverlayFade`.
- **Timing** — `OpenDelay` (pulse), `AutoCloseDelay`, `MouseOutTimeout`,
  `AnimationDuration`; `KeepOpenWhileActive` resets the auto-close on interaction so
  a complex interaction holds the cover open.
- **Styling** — `OverlayColor` (sanitized), `OverlayCssClass`, `OverlayTemplate`,
  `CssUrl`.
- **Accessible** — guard is a keyboard-operable `role="button"` (Enter/Space),
  content is `inert` while closed (with a no-`inert` fallback), focus management,
  `aria-expanded`, reduced-motion support.

## Files

- Control: `TSafetyCover.php`, `TSafetyCoverEffect.php`, `TSafetyCoverDirection.php`
- Client: `source/prado/controls/safetycover.js`, `assets/safetycover.css`
- Registered in `classes.php` and `packages.php`.

## Tests

- Unit: `tests/unit/Web/UI/WebControls/TSafetyCoverTest.php`
- vitest: `tests/js/controls/safetycover.test.js`
- Playwright: main, direction×fade matrix, accessibility, keep-open specs

Full check suite green (php-cs-fixer, phpstan, phpunit, vitest, Playwright).

PS.  The second example is too fast, and not a good example.  The Agent decided on that speed.  At least it works.
Also, I was hoping for a red covering with alpha, which is the default.


belisoful:
We can rename this.  I am not tied to it.

ctrlaltca:
Sorry, i missed the PR went from draft to "ready to merge"

PR #1253, opened 2026-08-29
Upgrades JS CallbackControl.onPostback - better logging on exception

belisoful:
## Callback dispatch: log on failure, fall back to the default action

`CallbackControl.onPostBack` (and `TActiveImageButton`) previously called `dispatch()` then `preventDefault()`. If `dispatch()` threw — typically from a `ClientSide` hook — `preventDefault()` never ran, so the failure was silent and the click still fell through to a postback.

Now the dispatch is wrapped: on throw, the error is logged via `Logger.error` and rethrown, and `preventDefault()` is deliberately skipped so a submit button degrades to a full-page postback instead of dead-ending. `TActiveImageButton` keeps its x/y coordinate inputs on failure so the fallback submit carries the click position.

JS-only; no PHP change. Adds unit coverage in `activecontrols.test.js` and a functional spec (`TActiveButtonPostBackFallbackTestCase`).

### DIFF
--- a/framework/Web/Javascripts/source/prado/activecontrols/activecontrols3.js
+++ b/framework/Web/Javascripts/source/prado/activecontrols/activecontrols3.js
@@ -6,9 +6,20 @@
 Prado.WebUI.CallbackControl = Prado.Class(Prado.WebUI.PostBackControl,
 {
 	onPostBack(options, event) {
-		const request = new Prado.CallbackRequest(options.EventTarget, options);
-		request.dispatch();
-		event.preventDefault();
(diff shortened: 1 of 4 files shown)


belisoful:
The Agents identified that if there is an exception, that "preventDefault" is not called and falls back to posting the page.  Any exception is written to the JS log as a matter of the browser function...  but it would immediately be erased by the post response.  Any intrinsically logged exceptions are unknown and unrecoverable.

This PR makes the logging of the exception deliberate and known.

The agents recommended rearranging the order of operation, but after considering the pros and cons, the way it is works but needs better exception handling.

PR #1254, opened 2026-08-30
WebControl Accessibility Audit Fixes

belisoful:
I ran an audit of the WebControls and found more than a few serious issues with accessibility.

The ActiveControls (In-Place too) and their parent controls were fixed up.
then the rest of the WebControls were the natural follow up.

[x] ActiveControls, In-Place, and their non-active Parents.
[x] all the rest of the non-active controls.

PS.  This came about because of an accessibility audit on the [some new] In-Place controls and found lots of issues in general too.

### DIFF
--- a/framework/Web/UI/WebControls/TTabView.php
+++ b/framework/Web/UI/WebControls/TTabView.php
@@ -61,6 +61,13 @@ class TTabView extends \Prado\Web\UI\WebControls\TWebControl
 		parent::addAttributesToRender($writer);
 
 		$writer->addAttribute('id', $this->getClientID());
+		// The view body is the panel controlled by its tab. Real navigation
+		// tabs (NavigateUrl) are plain links, not part of the tab pattern.
+		if ($this->getNavigateUrl() === '') {
+			$writer->addAttribute('role', 'tabpanel');
+			$writer->addAttribute('aria-labelledby', $this->getClientID() . '_0');
+			$writer->addAttribute('tabindex', '0');
+		}
 	}
 
 	/**
@@ -153,6 +160,15 @@ class TTabView extends \Prado\Web\UI\WebControls\TWebControl
 		if ($this->getVisible(false) && $this->getPage()->getClientSupportsJavaScript()) {
 			$writer->addAttribute('id', $this->getClientID() . '_0');
 
+			// A JS-switching tab is the focusable `tab` for its panel; roving
+			// tabindex keeps only the selected tab in the sequential tab order.
+			if ($this->getNavigateUrl() === '') {
+				$writer->addAttribute('role', 'tab');
+				$writer->addAttribute('aria-selected', $this->getActive() ? 'true' : 'false');
+				$writer->addAttribute('aria-controls', $this->getClientID());
+				$writer->addAttribute('tabindex', $this->getActive() ? '0' : '-1');
+			}
+
 			$style = $this->getActive() ? $this->getParent()->getActiveTabStyle() : $this->getParent()->getTabStyle();
 			$style->addAttributesToRender($writer);
 
@@ -165,18 +181,20 @@ class TTabView extends \Prado\Web\UI\WebControls\TWebControl
 	}
 
 	/**
-	 * Renders the content in the tab.
-	 * By default, a hyperlink is displayed.
+	 * Renders the content in the tab. A navigation tab ({@see setNavigateUrl
+	 * NavigateUrl}) renders a hyperlink; a JS-switching tab renders its caption
+	 * as text, so the `role="tab"` element itself is the only focusable control.
 	 * @param \Prado\Web\UI\THtmlWriter $writer the HTML writer
 	 */
 	protected function renderTabContent($writer)
 	{
-		if (($url = $this->getNavigateUrl()) === '') {
-			$url = 'javascript://';
-		}
 		if (($caption = $this->getCaption()) === '') {
 			$caption = '&nbsp;';
 		}
-		$writer->write("<a href=\"{$url}\">{$caption}</a>");
+		if (($url = $this->getNavigateUrl()) === '') {
+			$writer->write($caption);
+		} else {
+			$writer->write("<a href=\"{$url}\">{$caption}</a>");
+		}
 	}
 }
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -116,6 +116,7 @@ Docblocks inform and describe; it is not persuasive writing.
 - A full check consists of the 4 checks (in order): `php -l` compile, php-cs-fixer, phpstan, phpunit (all checks must pass successfully)
 - A full check must be done for code to be ready for git commit.
 - The per directory "<dir_path>/" information is found at "agents/<dir_path>/INDEX.md" to keep the framework uncluttered. 
+- All new and changed WebControls must be audited for accessibility at least once during development.
 - **The current version is 4.3.3. The next release version is 4.4.0**.
 
 ### ActiveControls JavaScript
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -129,6 +129,7 @@ TApplication
 - Method Doc Blocks must be **tight**, and have at minimum one sentence in the description.
 - Documentation additions/changes/removals should be integrated into the whole, at each level (of detail).
 - The per directory "<dir_path>/CLAUDE.md" is found at "agents/<dir_path>/INDEX.md" to keep the framework uncluttered.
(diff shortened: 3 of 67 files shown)


framework/Web/UI/WebControls/TTabView.php
```diff
@@ -165,18 +181,20 @@ public function renderTab($writer)
...
 	 */
 	protected function renderTabContent($writer)
 	{
-		if (($url = $this->getNavigateUrl()) === '') {
-			$url = 'javascript://';
-		}
 		if (($caption = $this->getCaption()) === '') {
 			$caption = '&nbsp;';
 		}
-		$writer->write("<a href=\"{$url}\">{$caption}</a>");
```
ctrlaltca, line 180: Unfortunately this is a breaking change we already made in the past. Removing the <a> from the tab will  break custom css targeting the tab captions. That's why the 'javascript://' fake url workaround was introduced.

ctrlaltca:
Looks really good, thank you! A small note on a TTabView change.. maybe it's time to just drop the workaround and make people update their css. Your choice

framework/Web/UI/WebControls/TTabView.php
belisoful: We do need to encode the rule regarding "javascript://"-urls into the agent files.  Obviously, the first instinct is to revert the url links for another solution. But depending on the research, you may be correct about dropping the workaround.

Accessibility is HUGE, especially for a framework.

belisoful:
I'll do some more research on options regarding the fake js url.  There will be some changes to this PR, eg we may upgrade this PR to addressing the js urls, but maybe not.

Basically, this aspect of PRADO probably needs a review given how far, how standardized, and how ubiquitous  browser technology has become.

belisoful:
# TTabView tab-caption anchor: analysis & choices (PR #1254)

## Background

The tab bar renders each tab as a styled `<div>`. Historically the caption was wrapped in an anchor with a fake URL:

**master (before this PR):**
```html
<div class="tab-normal" id="ctlX_0"><a href="javascript://">Caption</a></div>
```

**this branch (accessibility PR):**
```html
<div class="tab-normal" id="ctlX_0" role="tab" aria-selected="false"
     aria-controls="ctlX" tabindex="-1">Caption</div>
```

The tab-switching JS (`tabpanel.js`) binds click/keydown to the **outer div** by its `_0` id. It never used the anchor.

## Why the fake `javascript://` link existed

`javascript://` is a no-op URL (`//` is a JS line comment, so activating it does nothing). It existed only to render a real `<a>` element so that:

1. the caption was keyboard-focusable (anchors with `href` are in the tab order), and
2. custom CSS targeting `.tab a` kept working.

The actual switching was always done by the JS click handler, not the link. Comment 1 on the PR notes this anchor was removed once before, broke CSS, and the `javascript://` workaround was introduced to restore the anchor.

## The concrete problem this change causes

PRADO's own shipped default `framework/Web/UI/WebControls/assets/tabpanel.css` styles captions **through the anchor**:

```css
.tab-active a       { color: black; text-decoration: none; font-weight: bold; }
.tab-normal a       { color: gray;  text-decoration: none; font-weight: bold; }
.tab-normal a:hover,
.tab-normal a:focus { color: rgb(120,172,255); ... }
```

`tabpanel.css` is **unchanged on this branch**, so dropping the anchor already regresses PRADO's own default look (captions lose color, bold, hover/focus). Any user CSS using `.tab a` breaks the same way.

Second, self-inflicted regression: the old `:focus` color lived on the anchor and nothing replaced it on the div, so keyboard focus is now **invisible** — which undercuts the accessibility goal. This must be fixed regardless of which option is chosen.

## CSS migration (if the anchor is dropped)

A one-level selector move, mechanical to document:

```css
/* before        →  after */
.tab-active a       →  .tab-active
.tab-normal a       →  .tab-normal
.tab-normal a:hover,
.tab-normal a:focus →  .tab-normal:hover, .tab-normal:focus
```

## Can PRADO fix the JS or CSS?

- **JS**: nothing needed — it binds to the div, not the anchor, and is already unobtrusive (no inline handlers).
- **CSS**: PRADO must update its own `tabpanel.css` either way. It cannot rewrite users' CSS, but it can choose markup that keeps their CSS working (Options B/C).

## Modern alternatives considered

- **Server-rendered `onclick`**: a step backward — inline handlers are what strict CSP (`script-src`) blocks; PRADO already uses the cleaner attach-by-id model. Do not reintroduce inline JS.
- **`<button type="button" role="tab">`**: semantically fine, natively focusable, but breaks CSS worse (`.tab a` → `.tab button`) and fights the browser button reset. No advantage over a plain `role="tab"` div here.
- The correct modern tab pattern (`role=tablist/tab/tabpanel` + roving tabindex + arrow keys) is already implemented; the only open question is the anchor-for-CSS decision.

## Choices

| Option | Markup | CSS impact | `javascript://` gone? | a11y | Work |
|---|---|---|---|---|---|
| **A. Drop anchor** (current PR) | `<div role="tab">Caption</div>` | Breaks `.tab a`; update shipped CSS + doc migration | Yes | Best | update css + changelog |
| **B. Inert anchor shim** | `<div role="tab"><a>Caption</a></div>` (no href) | Keeps color/weight/hover; only `:focus` needs a new `.tab:focus` rule | Yes | Good (role on div; anchor inert) | tiny |
| **C. Anchor is the tab** | `<a role="tab" tabindex=...>Caption</a>` (no href), JS drives the anchor | Preserves all `.tab a` incl `:focus` | Yes | Good (tab role on nested anchor — slightly less ideal for tablist→tab) | JS rework |
| **D. Status quo + role** | keep `href="javascript://"`, add role | Zero CSS change | No — CSP-hostile link-to-nowhere | OK | trivial |

## Recommendation

**Option A.** Cleanest markup, correct ARIA pattern, removes the CSP-hostile fake URL, trivial mechanical migration. It is a real but small breaking change — appropriate for a 4.4.0 minor release framed as an accessibility fix, and it is what comment 2 invited ("maybe it's time to just drop the workaround... your choice"). Pair it with: update `tabpanel.css` to target the tab div, add a visible `:focus` style (fixing the focus regression), and add an UPGRADE/changelog note with the selector migration.

If breaking no one's CSS is required, **Option B** is the pragmatic compromise: removes the fake URL and keeps a11y while leaving virtually all existing CSS working (only a one-line `:focus` addition).

belisoful:
The above comment is an analysis for those not familiar with the issue referenced by @ctrlaltca.

The main purpose was to see what the options were.  in terms of a11y, Option A is the best and I am leaning towards that.   If this was 4.3.x, I'd choose Option C.   It was already stated that BC can bite the bullet.  Now that it's documented what the options and changes could be, is there a preference?

I am leaning heavy into accessibility because that is so important as the main weight factor in my preference.

ctrlaltca:
Go for option A :+1:

belisoful:
## Decision on the TTabView caption anchor: going with Option A

Thanks for the context on the history here — the note about the earlier BC break and why the `javascript://` workaround was introduced was really helpful, as was the suggestion that it may be time to just drop the workaround. After weighing it, I've gone with **dropping the anchor and updating the CSS**, and documenting the break.

### What changed
- A JavaScript-switching tab (a tab **without** a `NavigateUrl`) now renders its caption as **plain text on the `role="tab"` element**, instead of wrapping it in `<a href="javascript://">…</a>`.
- Tabs that **do** set `NavigateUrl` still render a real `<a href="…">` — no change there.
- The fake `javascript://` no-op link is gone.

### Why
- It makes the tab bar follow the WAI-ARIA tabs pattern properly: the `role="tab"` element is the focusable control (with roving `tabindex` + arrow-key navigation), rather than a link-to-nowhere carrying a tab role.
- `javascript://` is a no-op that strict Content-Security-Policy rejects; removing it is a small correctness win on its own.

### Backward compatibility - Changes to CSS
This is a breaking change **only** for custom CSS that styled the caption *through the anchor*. The fix is a one-level selector move:

| Before | After |
| --- | --- |
| `.tab-normal a` | `.tab-normal` |
| `.tab-active a` | `.tab-active` |
| `.tab-normal a:hover` | `.tab-normal:hover` |

### Framework changes included
- The bundled default `tabpanel.css` has been retargeted at the tab element accordingly (so PRADO's own default styling is unaffected).
- Added a visible `:focus-visible` outline on the focused tab, which also fixes a keyboard-focus-visibility gap the anchor removal would otherwise leave.
- Documented the break and the selector migration in `UPGRADE.md` (Upgrading from v4.3.3).

Given this lands in a minor release (4.4.0) and is an accessibility correctness fix, a small, clearly-documented CSS break feels like the right trade.

Issue #451, opened 2013-09-14
Grouping ValidationGroups


ctrlaltca:
_From [ra...@bhatia.at](https://code.google.com/u/100479290254572504264/) on March 14, 2013 19:20:14_

Right now, i can use ValidationGroup="" to group different validators to be executed when a button is hit.

However, this grouping is done on a site-wide level and causes problems when writing something like a login portlet (TTemplateControl) which is displayed multiple times on a page.

My current solution is something like this to add Parent->getUniqueId() to the ValidationGroup Attribute:

LoginPortlet.tpl

> <com:TRequiredFieldValidator
>     ValidationGroup="validateLogin<&#37;= $this->Parent->UniqueID &#37;>"
>     ControlToValidate="email"
>     Text="cannot be empty." />
> 
> <com:TActiveLinkButton ID="submitLoginButton" 
>     Text="login"
>     onCallback="loginSubmit"
>     ValidationGroup="validateLogin<&#37;= $this->Parent->UniqueID &#37;>" />

This works well for my purpose.

However, it would be nice to see such functionality inside the TBaseValidator implementation, e.g. via a GroupWithParent="true" attribute.

I do not fully know the policy for introducing such parameter (naming, implication of ViewState, etc.), thus asking for help via this issue :)

Thanks,
Raoul
PS. Thank you for your support during the last years and that you're constantly caring about such demanding users as me ;)

_Original issue: http://code.google.com/p/prado3/issues/detail?id=450_

belisoful:
I don't think this requires any action.

PR #1261, opened 2026-09-06
#450 Fix - Template attribute values mix ~, $, [, / tags with literal text

### LINKED ISSUE #450: Include additional text when localizing a validator's text
_From [ra...@bhatia.at](https://code.google.com/u/100479290254572504264/) on March 14, 2013 19:11:40_

I would like to add additional text/HTML code when localizing a validator's text. For example, i want to add a <br>

example 1 (working):
<com:TRequiredFieldValidator ControlToValidate="email" Text="<&#37;[Please enter your email address!]&#37;>" />

example 2 (not working):
<com:TRequiredFieldValidator ControlToValidate="email" Text="<&#37;[Please enter your email address!]&#37;><br>" />

This seems to be rooted in TTemplate::parseAttribute(), line 896:

> else if(preg_match('/\s_(<&#37;~._?&#37;>|<&#37;\$._?&#37;>|<&#37;\[._?\]&#37;>|<&#37;\/._?&#37;>)\s_/msS',$value,$matches) && $matches[0]===$value)

There, it correctly matches the localization tag but checks if the $matches[0] equals $value, which isn't true for the second example because:

$matches[0] is "<&#37;[Please enter your email address!]&#37;>" but
$value is "<&#37;[Please enter your email address!]&#37;><br>".

Is there a possibility to fix this?

Thanks,
Raoul

_Original issue: http://code.google.com/p/prado3/issues/detail?id=449_

belisoful:
TTemplate::parseAttribute() recognized <%~ %>, <%$ %>, <%[ ]%> and <%/ %> tags only when the tag was the entire attribute value. Anything else fell through to CONFIG_VALUE and the raw tag text was assigned to the property, so Text="<%[ Please enter your email! ]%><br>" set the literal string.

The whole-value typed check now runs first, preserving CONFIG_ASSET, CONFIG_PARAMETER and CONFIG_LOCALIZATION results for lone tags and the ID/SkinID validation that depends on them. The mixed-content branch, previously limited to <%= %> and <%# %>, accepts all six tag types and concatenates text and tag results into one CONFIG_EXPRESSION, or CONFIG_DATABIND when a <%# %> tag is present. The new protected helper parseAttributeTag() converts one tag to its PHP fragment and reuses parseExpression() for the ~, $, [ and / types. <%% %> statement tags stay literal in attributes.

Mixed values evaluate as auto-bind expressions in preRenderRecursive() with the template control as context.

- framework/Web/UI/TTemplate.php: parseAttribute() reorder and widen, parseAttributeTag() `@since` 4.4.0, class and method docblocks.
- tests/unit/Web/UI/TTemplateTest.php: 13 parseAttribute()/parseAttributeTag() tests including the reporter's template.
- tests/unit/Web/UI/TTemplateInstantiateInTest.php: 2 end-to-end tests asserting localized and parameter-substituted Text after auto-binding.
- agents/framework/Web/UI/TTemplate.md: full tag list, Attribute Values table, evaluation timing, new helper.

Backward compatible. Fixes #450.


belisoful:
a new base-line PRADO Template processor is added `<%! ......  %>`.  This acts almost like a `<%= ......  %>` except the binding is not in pre-render, but in first thing in TControl::initRecursive.

I'm not sure if this is the proper way to implement this yet.  I'm doing some more research on its implementation.  I'm looking into if it should be processed in the TTemplate or in the TPage onInit (or just after?).  This will determine if the bang processor has access to the page controls or not.  I'd like to have equivalency between the two operators (`!` and `=`).

The initial idea for `<%! expr %>` was something I was playing with ~5 years ago in my experimental branch.

belisoful:
## New `<%! %>` attribute tag (init binding)

This PR also adds a fourth attribute tag, `<%! expression %>`, alongside `<%= %>`, `<%# %>` and the typed tags.

### Why

`<%= %>` on a `TControl` is not applied when the template is instantiated. It registers an auto-binding that `preRenderRecursive()` evaluates, so the property is still empty during Init, Load, postback processing and validation. That makes it unusable for a class of properties:

- `ValidationGroup` / `CausesValidation`: validation runs in `raisePostBackEvent`, before PreRender, so a `<%= %>` group is never seen by the validator.
- Input defaults such as `Text="<%= $this->User->Email %>"` on a `TTextBox`: the PreRender evaluation overwrites whatever the user posted.
- Anything a control reads in `createChildControls()` or `onInit`.

The framework already special-cases `ID` and `SkinID` to evaluate immediately for exactly this reason. `<%! %>` generalizes that to any property. (It originates from a `<%- %>` tag in an old private Prado fork; the character was changed to `!` because `-` already means unescaped output in EJS and whitespace trimming in ERB, and `!` reads as "now".)

### Timing

| Tag | Mechanism | Applied |
|---|---|---|
| `<%# %>` | `bindProperty()` | `dataBind()` |
| `<%! %>` | `initBindProperty()` (new) | first line of the control's `initRecursive()`, once |
| `<%= %>` | `autoBindProperty()` | `preRenderRecursive()` |

`<%! %>` runs before the control's own `createChildControls()`, before theme skins, before `onInit`, and before `loadPageState` / `processPostData`. So on a postback, viewstate and posted data override the init value, which is the same behavior as setting the property in `onInit`. Controls added after Init get the binding applied through the existing lifecycle catch-up in `addedControl()`. Non-control `TComponent` targets have no lifecycle and are set at instantiation, the same fallback `<%= %>` already uses.

### Visibility rule

When a `<%! %>` expression runs, the whole template has been instantiated and attached, so every control in it is discoverable by ID (`$this->SomeId`) with `Parent`, `Page`, `NamingContainer`, `ClientID` and `UniqueID` resolved. A later sibling's own `<%! %>` values are not yet applied, because its `initRecursive()` has not run. Expression context is the template control, as with the other tags.

### Rules

- Attribute-only. In body content `<%! %>` is left as literal text; there is no init-time mechanism for `TCompositeLiteral`.
- A property takes exactly one tag type. Declaring it twice throws `template_property_duplicated`. Tags mixed in one value collapse to a single type with precedence `#` → `!` → `=`, so `Text="<%! 'a' %> <%= 10 %>"` is one init expression producing `a 10`.
- `ID` and `SkinID` accept `<%! %>` and keep their immediate evaluation.
- Values that must be recomputed after viewstate loads, or that depend on `onLoad` state, stay `<%= %>`.

### Examples

```html
<com:TButton Text="Save" ValidationGroup="<%! $this->ClientID . '_save' %>" />
<com:TRequiredFieldValidator ControlToValidate="Email" ValidationGroup="<%! $this->ClientID . '_save' %>" />
<com:TTextBox ID="Email" Text="<%! $this->User->Email %>" />
<com:TLabel Text="For <%! $this->Email->UniqueID %>" />
```

### Bugs fixed along the way

**Template ID tags evaluated twice.** `ID="<%= ... %>"` and `ID="<%$ ... %>"` on a `TControl` were evaluated for `registerObject()` but left typed as expressions, so `configureProperty()` evaluated them a second time: `<%= %>` registered an auto-binding on ID that threw at PreRender, and `<%$ %>` looked the evaluated string up as a parameter again. The property is now retyped to `CONFIG_VALUE` after evaluation. Regression tests cover both.

**`TControl::getPluginModule()` always returned `null`.** Since 4.2.0 the `_pluginmodule` field was declared without an initializer, so it started as `null`, while the getter only runs the module search when the field is `false` and uses `null` to mean "searched, none found". The search never ran. The field now initializes to `false`, documented as `null|false|IPluginModule`. The getter's docblock is `@return ?\Prado\Util\IPluginModule` for readers and IDEs with `@phpstan-return mixed` beneath it, so user code can call its own plugin module's methods on the result without PHPStan reporting them as undefined on the interface (verified with a level 8 probe). Tests cover a matching module, a non-matching module, and the once-only search.

### Surface

- `TTemplate::CONFIG_INIT_EXPRESSION`
- `TControl::initBindProperty()`, `TControl::initDataBindProperties()`, `TControl::RF_INIT_BINDINGS`
- `TControl::getPluginModule()` fix and docblock
- Docs in the `TTemplate` class docblock, `agents/framework/Web/UI/TTemplate.md` and `TControl.md`, and HISTORY.

belisoful:
This PR has 3 related fixes.
 - Add other `<%?  expr %>` into string computations in the TTemplateManager.
 - Add `<%!  expr  %>` for onInitRecursive property setter (like a default value).  This adds a new init binding to TControl.
 - Fixes a long but super simple bug regarding discovery of TControl::PluginManager.

belisoful:
Each of the edits-fixes are broken down into their own commits for their own examination.

ctrlaltca:
> Each of the edits-fixes are broken down into their own commits for their own examination.

Yep, thank you for that. I completed the code review, now i'm testing the change i a couple of big apps to detect any regression.. please be patient for another day, if everything is fine i'll merge it straight away

belisoful:
No issue on the delay.  We are beginning to move from the lower hanging fruit to the middle tree fruit in regard to fixes and enhancements.  Still plenty of low hanging fruit, but it is thinning.

belisoful:
Merging this PR may help #451 with the `ValidationGroup`.  _may_....   Just pointing this out for any agent reading this.

ctrlaltca:
After a couple of days of testing, no regression reported.

PR #1265, opened 2026-09-11
TUrlMappingPattern query string matching (Resolves #206)

### LINKED ISSUE #206: Feature Request: Add support for fully qualified URL-matching in TUrlMapping
_From [google...@pcforum.hu](https://code.google.com/u/112363926660835410968/) on October 20, 2009 04:03:18_

TUrlMapping (well, actually TUrlMappingPattern) currently only supports 
mapping based on the "pathinfo" part of the url, and provides no way to 
provide any mapping which also considers any parameters passed to the 
script in the "querystring" part of the url. I however fail to see why 
this limitation is there. Without this, one must fall back to 
use .htaccess rewrite techniques, contradicting the primary purpose of 
Prado's url mapping (which in my understanding is allowing developers to 
avoid those url rewrites from config, and moving that logic integral to 
the application), in cases where for ex. you want to have calls diverted 
to different pages based on specific query parameter values in the url. 
(For ex. you might have moved the functionality once implemented in the 
same page now apart, to two more pages, without invalidating already 
created and stored urls in the application, in search engines, etc.)

This limitation, however, could be easily overcomed would 
TUrlMappingPattern provide a way to use the result of $request->getRequestUri() instead of $request->getPathInfo(). Therefore I propose to add a new property to the latter class, which would enable developers to make use of that extra part of the url, too, in match-checks (againt the pathinfo+querystring part of the url), while also preserving backwards compatibility (matching against only the path info).

_Original issue: http://code.google.com/p/prado3/issues/detail?id=205_

belisoful:
Patterns could only match the PATH_INFO. Two ways now match the query string:

- `UrlMatchMode="Full"` matches the pattern, the text after its first '?' included, against the PATH_INFO and the query string. Defaults to PathInfo, the previous behavior. New TUrlMappingPatternUrlMatchMode enumerable.
- `query.<name>="<regex>"` constrains single GET variables, whatever their order in the URL. The whole value must match.

Both apply to construction as well: a pattern is used by constructUrl() only when the GET items satisfy its constraints, and extra variables are appended with '&' when the pattern already carries a query string.

## The Second Commit - TUrlMappingPattern Verbs negation never matched

A pattern with Verbs="!DELETE" matched no request at all. The gate
rejected any method not listed literally, so a negation could only fire
for a method that was both included and excluded, which is
contradictory. Only positive allow-lists worked, while the class
docblock documented the negated forms.

Matching moves to TUrlMappingPattern::matchesVerb(). A list of negated
verbs alone matches every method it does not exclude, a mixed list
matches a method that is included and not excluded, and the comparison
ignores case.

Fixes #206


belisoful:
should `query` constrain `post` also?
should we have `get.<name>="<regex>"`, `post.<name>="<regex>"`, `delete.<name>="<regex>"`, etc....?
or `query.get.<name>="regex"`, `query.post.<name>="regex"`, `query.delete.<name>="regex"`, etc....?

Issue #120, opened 2013-09-14
Add types in a soap service without property declaration


ctrlaltca:
_From [han...@gmail.com](https://code.google.com/u/101979232340910883863/) on February 16, 2009 00:15:26_

Hi , 
here a snippet to add complex type definition in the wsdl file. It could be 
useful where you have some methods returns mixed results.

```diff
# Index: WsdlGenerator.php
--- WsdlGenerator.php   ( revision 242 )
+++ WsdlGenerator.php   (working copy)
@@ -97,6 +97,18 @@
        $this->wsdlDocument = new Wsdl($className, $serviceUri, 
$encoding);

    $classReflect = new ReflectionClass($className);
-   
-       $comment = $classReflect->getDocComment();
-       if(strpos($comment, '@soaptype') !== false)
-       {
-           if (preg_match('/@soaptype\s+(\w+([])?)/mi', $comment, $match)) {
-               $param = array();
-               $param['type'] = $this->convertType($match[1]);
-               $param['name'] = $match[1];
-               $this->types[$match[1]][] = $param;
-           }
-       }  
- 
  $methods = $classReflect->getMethods();
  
  foreach ($methods as $method) {
```
_Original issue: http://code.google.com/p/prado3/issues/detail?id=119_

belisoful:
Implemented, 17 years on. One note on where: `WsdlGenerator` no longer lives in this repository — it was split out to [pradosoft/prado-wsdlgenerator](https://github.com/pradosoft/prado-wsdlgenerator), which Prado pulls in as `pradosoft/prado-wsdlgenerator: ^1`. The change is in a PR there and ships in 1.2.0. Prado picks it up with no constraint change.

## The tag

`@soaptype` goes on the provider class and declares a complex type, so a type reaches the WSDL even when no `@soapmethod` signature names it — which is what a method returning mixed results needs in order to tell the client every shape it can return:

```php
/**
 * @soaptype Person
 * @soaptype Address[]
 */
class MyProvider
{
	/**
	 * Look up a record.
	 * @param string $key the key to look up
	 * @return mixed a Person or an array of Address
	 * @soapmethod
	 */
	public function lookup($key) { /* ... */ }
}
```

The tag takes a class name, optionally suffixed with `[]` for the array form, which also declares the element type. It runs through the existing `convertType()`, so the declared type pulls in its own `@soapproperty` elements recursively. A class carrying no `@soaptype` generates byte-identical WSDL to 1.1.

[...]

PR #1266, opened 2026-09-16
ENH: TSoapServer::WsdlStyle selects the style of a generated WSDL

belisoful:
A generated WSDL was always rpc with SOAP encoding, which WS-I Basic Profile 1.1 prohibits, and nothing could ask for anything else.

Add the WsdlStyle property, 'rpc' by default and 'document' for the profile. Being a writable property of TSoapServer, it is settable as an attribute of the <soap> element. It has no effect when WsdlUri names a WSDL to serve, and the style joins the WSDL cache key, so a document generated under one style is never served for the other.

'document' needs pradosoft/prado-wsdlgenerator 1.2, and the composer requirement stays at ^1 because the property degrades rather than depending on it. The style names its own constants instead of the ones the package added, so a server still constructs against an earlier generator, and the default style is passed the way it always was, so that generator still serves it. Any other style is looked for before it is passed: PHP discards an argument a function does not declare, so a generator that predates the style would otherwise answer rpc without complaint. It is refused instead, naming the version it needs.

This as an enhancement from #120.


belisoful:
The first commit in the style enhancement and the second/third is the bug fix for php config #1267

PR #1268, opened 2026-09-16
Misc test and code fixes; try enabling tests for php 8.4 and php 8.5

ctrlaltca:
Workaround new warnings about NaN introduced by php 8.5
The bcmath and bz2 extensions are needed by tests, add them in require-dev
Fix locale-related tests when php default locale is not en_US
Try enabling tests for php 8.4 and php 8.5


belisoful:
time to upgrade my prado repos' unit tests and pradosoft's other repos too.  I'll key from this.  Not ASAP, but as we evolve.

PR #1270, opened 2026-09-16
move phpstan check to php 8.5 and switch to level 3

ctrlaltca:
Apart from the phpdoc changes and some minor code modification to ensure the correct data type is returned, a few major code modifications stand out:
* all __set() magic methods are now returning void, as per https://www.php.net/manual/en/language.oop5.overloading.php#object.set: _The return value of [__set()](https://www.php.net/manual/en/language.oop5.overloading.php#object.set) is ignored_
* TTemplate related classes now explicitly checks for `getDirective` and `getIsSourceTemplate` since ITemplate doesn't declare those method in the interface
* fixed a bug in `explode` call at https://github.com/pradosoft/prado/pull/1270/changes#diff-b347c513e2e340aad0c15161cc5c3b5984642bda9a66804e70327b90133f1b6aL96
* fixed a possible bug in  https://github.com/pradosoft/prado/pull/1270/changes#diff-a3a895a492e380ca3fab341f4da207a7b1c80ec91a7c4616dea985161b05ede3L1825


framework/Web/UI/WebControls/TWizard.php
```diff
@@ -187,11 +187,11 @@ public function addParsedObject($object)
 	}
 
 	/**
-	 * @return TWizardStep the currently active wizard step
+	 * @return ?TWizardStep the currently active wizard step
 	 */
 	public function getActiveStep()
 	{
-		return $this->getMultiView()->getActiveView();
+		return $this->getWizardSteps()->itemAt($this->getActiveStepIndex());
```
belisoful: This will throw when no step is active.  Value `-1`

Suggested:
```php
/** @return ?TWizardStep the currently active wizard step, null if none */
public function getActiveStep()
{
	$step = $this->getMultiView()->getActiveView();
	return $step instanceof TWizardStep ? $step : null;
}
```

framework/Web/UI/TTemplate.php
```diff
@@ -258,7 +258,7 @@ public function instantiateIn($tplControl, $parentControl = null)
 		if ($parentControl === null) {
 			$parentControl = $tplControl;
 		}
-		if (($page = $tplControl->getPage()) === null && ($service = $this->getService()) !== null && $service->isa(TPageService::class)) {
+		if (($page = $tplControl->getPage()) === null && ($service = $this->getService()) !== null && $service instanceof TPageService) {
```
belisoful: This change removes the ability for an alternative service to mimic a TPageService.

`isa` is Prado's behavior-aware version of PHP's `instanceof`.   It is slower, so that's the trade-off on a hot path.

framework/Util/TUtf8Converter.php
```diff
@@ -93,7 +93,7 @@ public static function fromUTF8($string, $to, $lang = null)
 	public static function parseEncodingLanguage(string &$encoding, &$lang)
 	{
 		if (strpos($encoding, '.') !== false) {
-			$parts = explode($encoding, '.', 1);
+			$parts = explode('.', $encoding, 2);
```
belisoful, line 96: Take it or leave it, there is no unit test for this class. This could be the start of testing this class.

If an agent is involved, a full coverage green unit test is a prompt away these days.

framework/I18N/TGlobalization.php
```diff
@@ -150,15 +150,15 @@ protected function setAppGlobalization()
 	}
 
 	/**
-	 * @return string default culture
+	 * @return bool wether to translate from default culture
```
belisoful: SP "whether".

framework/I18N/TGlobalization.php
```diff
@@ -150,15 +150,15 @@ protected function setAppGlobalization()
...
 	 */
 	public function getTranslateDefaultCulture()
 	{
 		return $this->_translateDefaultCulture;
 	}
 
 	/**
-	 * @param bool $value default culture, e.g. <tt>en_US</tt> for American English
+	 * @param bool $value wether to translate from default culture
```
belisoful: SP "whether".

framework/Shell/TShellAction.php
```diff
@@ -31,11 +31,12 @@ abstract class TShellAction extends \Prado\TComponent
...
 
 	/**
-	 * @return TShellApplication current application instance
+	 * @return ?TShellApplication current application instance
 	 */
 	public function getApplication()
 	{
-		return Prado::getApplication();
+		$application = Prado::getApplication();
+		return $application instanceof TShellApplication ? $application : null;
```
belisoful, line 39: I like this alignment.

But THelpAction dereferences self->getApplication and needs a null guard over there.

framework/TComponent.php
```diff
@@ -2051,7 +2056,7 @@ protected static function instanceBehavior($behavior)
 			$behavior = Prado::createComponent($behavior);
 			$init = true;
 		}
-		if (!($behavior instanceof IBaseBehavior)) {
+		if (!($behavior instanceof IBaseBehavior) || !($behavior instanceof TComponent)) {
```
belisoful: The TComponent check is quite appropriate.

It would be useful to break out the exception from the IBaseBehavior with its own exception message.

phpstan.neon.dist
```diff
@@ -1,6 +1,6 @@
 parameters:
-    phpVersion: 80300 # PHP 8.3
-    level: 2
+    phpVersion: 80500 # PHP 8.5
```
belisoful: This can take a range and it will validate each PHP version.  We already target PHP 8.1 through 8.5 (at this point).

```yml
parameters:
    phpVersion:
        min: 80100
        max: 80500
    level: 3
```

framework/Web/UI/TTemplateControl.php
```diff
@@ -100,7 +100,7 @@ public function setTemplate($value)
 	public function getIsSourceTemplateControl()
 	{
 		if (($template = $this->getTemplate()) !== null) {
-			return $template->getIsSourceTemplate();
+			return method_exists($template, 'getIsSourceTemplate') && $template->getIsSourceTemplate();
```
belisoful: the `method_exists` problem is resolved in this PR at  `IDataMetaData` with the addition of `@method` tags in the doc block.

Put this
```php
/**
 * @method array getDirective() Returns the template directive settings.
 * @method bool getIsSourceTemplate() Returns whether the template is a source template.
 * @method string getContextPath() Returns the template context path.
*/
```
into ITemplate.

then all these series of `method_exists` can be removed.

belisoful: the six other ITemplate implementers are Jui and wizard templates that never reach those call sites. Declaring the methods as real interface members would be cleaner but would break those six classes and any user templates, so the tags are the right call for 4.4.0.

framework/Web/UI/TTemplateControlInheritable.php
```diff
@@ -105,11 +105,11 @@ public function doTemplateForClass($parentClass)
 	public function getIsSourceTemplateControl()
 	{
 		if (null !== ($_template = $this->getTemplate())) {
-			return $_template->getIsSourceTemplate();
+			return method_exists($_template, 'getIsSourceTemplate') && $_template->getIsSourceTemplate();
 		}
 
 		return ($_template = $this->getService()->getTemplateManager()->getTemplateByClassName(get_parent_class($this)))
-			? $_template->getIsSourceTemplate()
+			? method_exists($_template, 'getIsSourceTemplate') && $_template->getIsSourceTemplate()
```
belisoful: This looks like the last `method_exists` in the set from above.

framework/Web/UI/ActiveControls/TActiveHiddenField.php
```diff
@@ -21,7 +21,7 @@
  *
  * @author Carl G. Mathisen <carlgmathisen@gmail.com>
  * @since 3.1
- * @method TActiveControlAdapter getAdapter()
+ * @method TActiveControlAdapter<\Prado\Web\UI\ActiveControls\TBaseActiveCallbackControl> getAdapter()
```
belisoful: ` * @method TActiveControlAdapter<\Prado\Web\UI\ActiveControls\TBaseActiveControl> getAdapter()` is the more accurate base class.

framework/Web/UI/ActiveControls/TActiveHiddenField.php
belisoful: `	 * @return TBaseActiveControl standard active control options.` is more honest and accurate.

belisoful: `getClientSide` apparently can be deleted after the return type above is corrected.

> I recommend deleting the whole method. It has thrown on every call since the commit that introduced the class, so no working code depends on it, and the removal does not need the deprecation cycle a minor release would otherwise want. If the team prefers to keep the symbol, the alternative is to leave it, tag it @deprecated 4.4.0, and have it throw a TNotSupportedException explicitly. The setValue() override at line 66 is fine as is, since canUpdateClientSide() lives on TBaseActiveControl.

framework/Web/UI/ActiveControls/TActiveDataGrid.php
```diff
@@ -44,7 +44,7 @@
  *
  * @author LANDWEHR Computer und Software GmbH <programmierung@landwehr-software.de>
  * @since 3.1.9
- * @method TActiveControlAdapter getAdapter()
+ * @method TActiveControlAdapter<\Prado\Web\UI\ActiveControls\TBaseActiveCallbackControl> getAdapter()
```
belisoful: ` * @method TActiveControlAdapter<\Prado\Web\UI\ActiveControls\TBaseActiveControl> getAdapter()` is more accurate.

framework/Web/UI/ActiveControls/TActiveDataGrid.php
belisoful: `	 * @return TBaseActiveControl standard active control options.` is more accurate.

belisoful:
Review Completed.

PRADO has some pre-8.4/8.5 method dependencies.  These are methods we depend on that were removed from PHP after 8.3.  there are (symfony) polyfills to make up the difference.  These older methods should be replaced or re-created for PRADO in PHP 8.5.

Also,  I'm looking at implementing or requiring a repo for PHP 8 features like Fibers, which we're then talking about Promises, Fulfillment, Cancellations, context switching, and a number of other things for async.  This is a very rich vein that I don't think we want to re-create.

I'll offer [`revolt/event-loop`](https://github.com/revoltphp/event-loop), [`amphp/amp`](https://github.com/amphp/amp), and other [repos of theirs](https://github.com/orgs/amphp/repositories).  We have our own PRADO layer on top of it.  Our IO Reactor Class is reformulated. Cron needs this.  Prado Application (User, etc) State needs to be saved and restored on a fiber context switch.   IO Http Client would need Fibers for asynchronous operation.   other blocking operations in Prado would be Fiber wrapped.

belisoful:
Also important to note, in general, that 8.1 security support was ended Dec. 31, 2025.  8.2 security support ends at the end of this year.  I don't think we are going to ship Prado v4.4 before the end of the year.  We've got a lot of great changes going on, but also a lot of settling to do.

I'm finding bugs in 4.3.3 from dev in v4.4 that does need to be back ported.  This should be easy enough. but the backlog is real.  Data is a big subsystem.  I'm still going on that, there are a lot of balls in the air right now.

ctrlaltca:
First thank you.. this has been possible only because of the extensive tests you added in the last months.

> Review Completed.

Can you approve the PR or ask for modifications?

> PRADO has some pre-8.4/8.5 method dependencies. These are methods we depend on that were removed from PHP after 8.3. there are (symfony) polyfills to make up the difference. These older methods should be replaced or re-created for PRADO in PHP 8.5.

Do you have a list? phpstan should report usage of deprecated methods but i have no mention here

> Also, I'm looking at implementing or requiring a repo for PHP 8 features like Fibers, which we're then talking about Promises, Fulfillment, Cancellations, context switching, and a number of other things for async. This is a very rich vein that I don't think we want to re-create.
> I'll offer [`revolt/event-loop`](https://github.com/revoltphp/event-loop), [`amphp/amp`](https://github.com/amphp/amp), and other [repos of theirs](https://github.com/orgs/amphp/repositories). We have our own PRADO layer on top of it. Our IO Reactor Class is reformulated. Cron needs this. Prado Application (User, etc) State needs to be saved and restored on a fiber context switch. IO Http Client would need Fibers for asynchronous operation. other blocking operations in Prado would be Fiber wrapped.


Definitely a good topic for future development

ctrlaltca:
> Also important to note, in general, that 8.1 security support was ended Dec. 31, 2025. 8.2 security support ends at the end of this year. I don't think we are going to ship Prado v4.4 before the end of the year. We've got a lot of great changes going on, but also a lot of settling to do.

Yes, probably targeting 8.3 is more realistic

belisoful:
It's been amazing to participate and contribute.   Thank you for the recognition.  It's easy to see what many of these  new things, updates, and fixes unlock.  I'm very happy the unlocks are working.

> Can you approve the PR or ask for modifications?

There are some pending review items that require some modifications. I hope they are showing up for you. It may be minor, there are two spelling errors in TGlobalization comments of the dozen item pending.  If the pending items are not clear, I'll gladly put more effort into describing the issue and the modification.

I'm doing the method review for PHP 8.3, 8.4, and 8.5 in regards to deprecation of 8.1 and 8.2....

ctrlaltca:
> There are some pending review items that require some modifications. I hope they are showing up for you. It may be minor, there are two spelling errors in TGlobalization comments of the dozen item pending. If the pending items are not clear, I'll gladly put more effort into describing the issue and the modification.

I just double checked but unfortunately i don't see any comment, the review status is still "pending"

belisoful:
I don't doubt that you aren't seeing them.  So then where is the difference coming from?  Here is a page shot of the TGlobalization review:
Screenshot 2026-09-17 at 1 33 21 AM

This is what I am seeing in the PR Change screen, with 2 review items in TGlobalization.

belisoful (changes requested):
Finished Review.

I am so sorry.  I didn't click the Submit button!   My humble apologies!!!    I seriously think that was my first PR Review on GitHub.

belisoful:
The agent was suggesting that Prado might be using deprecated or removed methods as PHP 8 progressed...   It was a probability that it had not verified before saying it.  I repeated it as a possibility.

It's not as important as the probability suggested.  Here is what it found.  

------

# PHP 8.1–8.5 deprecations and removals affecting PRADO

This issue lists the PHP functions, constants, and settings that PRADO calls and that PHP 8.1 through 8.5 deprecates or removes. It also lists any calls to PHP 8.2+ APIs that would need a polyfill while PRADO supports PHP 8.1.

**Scope:** `framework/` at [`0003b66`](https://github.com/pradosoft/prado/tree/0003b6653403e3b05dbb7576738aefa495d675cc). Comments, docblocks, and false positives are excluded. False positives include SQL backticks and PRADO's own `contains()`, `setValue()`, and `set()` methods. Line links are permalinks to that commit.

---

## 1. PHP 8.2+ APIs that need a polyfill

**None.** PRADO calls no function or class that is missing from PHP 8.1.

- **Functions:** none of the functions in `symfony/polyfill-php82` through `symfony/polyfill-php85` are called. Examples include `json_validate`, `array_find`, `array_any`, `array_all`, `mb_trim`, `mb_str_pad`, `str_increment`, `array_first`, `array_last`, and `get_error_handler`. The 8.2–8.5 additions that no polyfill covers are not called either. Examples include `ini_parse_quantity`, `http_get_last_response_headers`, `request_parse_body`, and `curl_multi_get_handles`.
- **Classes:** `Random\*`, `Dom\*`, `Pdo\*`, `Uri\*`, `BcMath\Number`, `RoundingMode`, and the `Date*Exception` classes are not used.
- **`#[\SensitiveParameter]` (PHP 8.2)** is the only 8.2+ feature in use. It marks 9 parameters in 8 files: `TDbCache`, `TDbConnection` (2), `IUserManager`, `TAuthManager`, `TDbUser`, `TDbUserManager`, `TUserManager`, and `TShellLoginBehavior`. It needs no polyfill. PHP 8.1 ignores the attribute, so those values are not redacted from stack traces on 8.1.
- `array_is_list()` is PHP 8.1, the current minimum.

A polyfill only adds missing new functions. It does not help with the deprecated calls in section 2, which need to be rewritten.

---

## 2. Deprecated or removed PHP APIs in use

### Hard deprecations

| Class::method | Line(s) | PHP API | Deprecated | Fix |
|---|---|---|---|---|
| `TCallbackErrorHandler::getExceptionStackTrace` | [112](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Web/UI/ActiveControls/TCallbackErrorHandler.php#L112) | `@strftime()` | 8.1 (removed in 9.0) | `date('Y-m-d H:i')` |
| `THttpSession::setCookieMode` | [375](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Web/THttpSession.php#L375), [378](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Web/THttpSession.php#L378) | INI `session.use_only_cookies = 0` (modes `None` and `Allow`) | 8.4 | Remove `None` and `Allow`. Both are already `@deprecated` since 4.3.1. |
| `THttpSession::getCookieMode` | [354](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Web/THttpSession.php#L354) | Reads `session.use_only_cookies === '0'` | 8.4 (the `0` value) | Remove the `Allow` branch together with the setter |
| `THttpSession::setUseTransparentSessionID` | [457](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Web/THttpSession.php#L457) | INI `session.use_trans_sid = 1` | 8.4 | Deprecate the property and accept only `false` |
| `THttpRequest::constructUrl` | [790–791](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Web/THttpRequest.php#L790-L791) | `SID` constant appended to URLs (GET-based sessions) | 8.4 | The branch is already skipped on 8.4+. Remove it when the minimum PHP version is 8.4. |
| `TPhpErrorException::__construct` | [47](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Exceptions/TPhpErrorException.php#L47) | `E_STRICT` constant | 8.4 | The branch is already skipped on 8.4+. Remove it when the minimum PHP version is 8.4. |
| `TFopenHttpClient::download` | [59](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/IO/HttpClient/TFopenHttpClient.php#L59), [65](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/IO/HttpClient/TFopenHttpClient.php#L65), [69](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/IO/HttpClient/TFopenHttpClient.php#L69) | `$http_response_header` magic variable | 8.5 | Use `http_get_last_response_headers()` on 8.4+ and the variable below 8.4 |
| `TCurlHttpClient::download` | [99](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/IO/HttpClient/TCurlHttpClient.php#L99) | `curl_close()` | 8.5 (no-op since 8.0) | Remove the call |
| `TEtcdCache::request` | [251](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Caching/TEtcdCache.php#L251) | `curl_close()` | 8.5 (no-op since 8.0) | Remove the call |
| `assets/captcha.php` (top-level script) | [126](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Web/UI/WebControls/assets/captcha.php#L126) | `imagedestroy()` | 8.5 (no-op since 8.0) | Remove the call |

### Soft deprecation: `__sleep()` / `__wakeup()` (PHP 8.5)

PHP 8.5 documents `__sleep()` and `__wakeup()` as deprecated in favor of `__serialize()` and `__unserialize()`. PHP 8.5 emits no runtime notice for them. `TComponent::__sleep()` is the base of PRADO's serialization model through `_getZappableSleepProps()`, so a migration has to keep that hook.

| Class | `__sleep` | `__wakeup` |
|---|---|---|
| `TComponent` | [2523](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/TComponent.php#L2523) | [520](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/TComponent.php#L520) |
| `TActiveRecord` | [211](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Data/ActiveRecord/TActiveRecord.php#L211) | [219](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Data/ActiveRecord/TActiveRecord.php#L219) |
| `TMappedStatement` | [887](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Data/SqlMap/Statements/TMappedStatement.php#L887) | [879](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Data/SqlMap/Statements/TMappedStatement.php#L879) |
| `TDbConnection` | [163](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Data/TDbConnection.php#L163) | |
| `TDbCommand` | [64](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Data/TDbCommand.php#L64) | |
| `TSqlMapStatement` | [304](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Data/SqlMap/Configuration/TSqlMapStatement.php#L304) | |
| `TParameterProperty` | [136](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Data/SqlMap/Configuration/TParameterProperty.php#L136) | |
| `TResultProperty` | [340](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Data/SqlMap/Configuration/TResultProperty.php#L340) | |
| `TPreparedStatement` | [63](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Data/SqlMap/Statements/TPreparedStatement.php#L63) | |
| `TSqlMapObjectCollectionTree` | [198](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Data/SqlMap/Statements/TSqlMapObjectCollectionTree.php#L198) | |
| `TJuiControlOptions` | [130](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Web/UI/JuiControls/TJuiControlOptions.php#L130) | |
| `TNull` | [200](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Collections/TNull.php#L200) | |
| `TStyle` | | [217](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Web/UI/WebControls/TStyle.php#L217) |
| `TWeakCallableCollection` | | [105](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Collections/TWeakCallableCollection.php#L105) |
| `TWeakList` | | [99](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Collections/TWeakList.php#L99) |
| `TWeakMap` | | [87](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Collections/TWeakMap.php#L87) |
| `TFnStream` | | [165](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/IO/Stream/TFnStream.php#L165) |
| `TLazyOpenStream` | | [142](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/IO/Stream/TLazyOpenStream.php#L142) |
| `TPumpStream` | | [171](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/IO/Stream/TPumpStream.php#L171) |
| `TPermissionsManagerPropertyTrait` | | [49](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Security/Permissions/TPermissionsManagerPropertyTrait.php#L49) |

### Related: weak random fallback

This is a security issue, not a deprecation. On PHP 8.2+, `random_bytes()` throws `\Random\RandomException`, as the `@todo` comments note. Both methods below catch that exception and fall back to `sha1(uniqid(mt_rand(), true))`, which is not cryptographically secure. The fallback should be removed so that the error propagates.

| Class::method | Line |
|---|---|
| `TSecurityManager::generateRandomKey` | [275](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Security/TSecurityManager.php#L275) |
| `TCaptcha::generateRandomKey` | [451](https://github.com/pradosoft/prado/blob/0003b6653403e3b05dbb7576738aefa495d675cc/framework/Web/UI/WebControls/TCaptcha.php#L451) |

---

## 3. Checked and not present in `framework/`

| PHP version | Deprecated API or construct |
|---|---|
| 8.1 | `date_sunrise` / `date_sunset`, `mhash*`, `FILTER_SANITIZE_STRING` / `FILTER_SANITIZE_STRIPPED`, `PDO::FETCH_SERIALIZE`, `key()` / `current()` / `next()` / `reset()` on objects (all uses operate on arrays) |
| 8.2 | `utf8_encode` / `utf8_decode`, `"${var}"` interpolation, the mbstring `QPrint` / `Base64` / `Uuencode` / `HTML-ENTITIES` encodings, `"self::"` / `"parent::"` / `"static::"` string callables |
| 8.3 | `get_class()` / `get_parent_class()` without arguments, `assert_options()` and `ASSERT_*`, `MT_RAND_PHP`, `U_MULTIBYTE_SUPPORT`, `NumberFormatter::TYPE_CURRENCY` |
| 8.4 | Implicitly nullable parameter types, `trigger_error()` with `E_USER_ERROR`, `lcg_value()`, `mysqli_ping` / `mysqli_kill` / `mysqli_refresh`, `session.sid_length` / `session.sid_bits_per_character`, `xml_set_object()`, the single-string `ReflectionMethod` constructor, `SOAP_FUNCTIONS_ALL`, `CURLOPT_BINARYTRANSFER` |
| 8.5 | `finfo_close()`, `xml_parser_free()`, `curl_share_close()`, `ReflectionProperty::setAccessible()` / `ReflectionMethod::setAccessible()`, the backtick operator, `(boolean)` / `(integer)` / `(double)` / `(binary)` casts, `SplObjectStorage::attach()` / `detach()` / `contains()` (PRADO's `contains()` calls are on its own collections), driver-specific `PDO::MYSQL_*` / `PDO::PGSQL_*` / `PDO::SQLITE_*` constants, `DATE_RFC7231` |

`tests/` has 29 `ReflectionProperty::setAccessible()` / `ReflectionMethod::setAccessible()` calls in 13 files. These calls do nothing since PHP 8.1 and are deprecated in 8.5.

framework/Web/UI/WebControls/TWizard.php
ctrlaltca: Fixed, thank you

framework/I18N/TGlobalization.php
ctrlaltca: Thank you, fixed alongside other 39 instances of the same error :goat:

framework/I18N/TGlobalization.php
ctrlaltca: Thank you, fixed alongside other 39 instances of the same error :goat:

phpstan.neon.dist
ctrlaltca: nice catch, thank you

framework/Web/UI/TTemplateControl.php
ctrlaltca: Fixed, thank you

framework/Web/UI/TTemplateControlInheritable.php
ctrlaltca: fixed, thank you

framework/Web/UI/ActiveControls/TActiveHiddenField.php
ctrlaltca: Nice catch, removed.

framework/Web/UI/ActiveControls/TActiveDataGrid.php
ctrlaltca: Correct, thank you. Fixed.

framework/Web/UI/ActiveControls/TActiveDataGrid.php
ctrlaltca: Correct, thank you. Fixed.

framework/Web/UI/ActiveControls/TActiveHiddenField.php
ctrlaltca: Correct, thank you. Fixed.

framework/Web/UI/ActiveControls/TActiveHiddenField.php
ctrlaltca: Correct, thank you. Fixed.

framework/Shell/TShellAction.php
ctrlaltca: I added a null guard to THelpAction, thank you

framework/TComponent.php
ctrlaltca: Yep, i added a separate throw for object_not_a_component

framework/Web/UI/TTemplate.php
ctrlaltca: I restored the `isa()` call and added a phpdoc comment to reassure phpstan about the datatype of $service

phpstan.neon.dist
ctrlaltca: I had to limit testing to php 8.4 and 8.5 since TStream* classes use [StreamBucket](https://www.php.net/manual/en/class.streambucket.php) and that's not available before php 8.4. Previous php versions used stdClass instead, so the datalen property had no strict type, causing it to be detected as float instead of int

ctrlaltca:
> The agent was suggesting that Prado might be using deprecated or removed methods as PHP 8 versions progressed...

This is outside the scope of this PR, that is already enough big to be hard to follow. I've created https://github.com/pradosoft/prado/issues/1272

phpstan.neon.dist
```diff
@@ -1,6 +1,8 @@
 parameters:
-    phpVersion: 80300 # PHP 8.3
-    level: 2
+    phpVersion:
+        min: 80400
```
belisoful: With min: 80100 PHPStan reports two errors, both the $consumed accumulation in TStreamCodecFilter.php:39 and TStreamFilter.php:196, where older stubs type the bucket length loosely. I tested casting the added value to int on those two lines with min: 80100, and the full framework analysis is clean. That is a two-line change and it makes the analyzer cover the three PHP versions the project still supports. If the team prefers to leave 8.1 to 8.3 unanalyzed, a comment in phpstan.neon.dist saying so would stop the next reader from asking.

framework/Util/TUtf8Converter.php
belisoful, line 96: Given this is public facing, a note about this fix in HISTORY.md may be warranted.

framework/Web/UI/ActiveControls/TActiveHiddenField.php
```diff
@@ -37,21 +37,13 @@ public function __construct()
...
 	 */
 	public function getActiveControl()
 	{
 		return $this->getAdapter()->getBaseActiveControl();
 	}
 
-	/**
-	 * @return TCallbackClientSide client side request options.
-	 */
-	public function getClientSide()
```
belisoful, line 50: Given this is public facing, a note about this removal in HISTORY.md may be warranted?  It may not be important enough.  Something to consider.  not required.

framework/Web/UI/TTemplate.php
```diff
@@ -258,7 +258,8 @@ public function instantiateIn($tplControl, $parentControl = null)
 		if ($parentControl === null) {
 			$parentControl = $tplControl;
 		}
-		if (($page = $tplControl->getPage()) === null && ($service = $this->getService()) !== null && $service->isa(TPageService::class)) {
+		if (($page = $tplControl->getPage()) === null && ($service = $this->getService()) !== null && $service instanceof TComponent && $service->isa(TPageService::class)) {
+			/** @var TPageService $service */
```
belisoful: This line is only required because the PHPStan extension isn't quite right.  This is a separate issue.  I'm looking at this right now.

This is not a change for you, but a note so you know. I'll remove things like this in the PHPStan PR later.
You can click "resolve" when read.

belisoful (changes requested):

belisoful:
@ctrlaltca 4 notes, 2 changes requested, and 1 admittance of personal responsibility #1274.  The PHP 8.x method post was removed for the issue created from it to keep this PR clean.

edit:  it's 3.  The TTemplate became a 3rd change with #1274

framework/Util/TUtf8Converter.php
belisoful: Keeping the HISTORY.md up to date is not something I've been strict about myself...  so, this is a self-reminder.  I'm going to go back and ensure that my own PRs are documented here (properly, of course).

Just PRed #1273 to update the HISTORY.md with recent changes.

framework/Web/UI/TTemplate.php
belisoful: #1274 fixes the need for the additional `@var` comment.  The comment can be removed

framework/Web/UI/ActiveControls/TActiveHiddenField.php
ctrlaltca: I think noone ever used this.

phpstan.neon.dist
ctrlaltca: Ok, I added the 2 casts and re-enable testing from 8.1

belisoful:
@ctrlaltca Not sure if you saw it, TPropertyValue has a conflict from the INF merge.  I was expecting this.   I'm circling back to see if it's ready to merge before heading to bed.   I thought it worth a comment to help expedite the PR if that is what you are looking for.

The TTemplate `@var` comment is not necessary with the new PHPStan fixes.   Also, my apologies for clobbering the composer.json suggestions you had added before.   it got lost is the large number of files that got updated with the name spacing of Tests.

framework/Web/UI/TTemplate.php
ctrlaltca: Fixed, thank you

ctrlaltca:
Ok, this is finally good to go now.

belisoful:
@ctrlaltca Cleared. OK to merge.   It's your merge message.

PR #1271, opened 2026-09-17
Namespacing Prado Tests, Unit Tests, and Test Harness Classes in dev with `Prado\Test\Unit\`

belisoful:
This puts the Prado Tests, Unit Tests, Test Harness Classes into namespaces on dev repos.

It makes some folder names PascalCase.
Break out combined class in files into their own files.


belisoful:
Waiting on pradosoft/prado-wsdlgenerator#5.  This PR will basically require a few lines of change to all WIP Testing classes.

It's a good thing we're doing this too. Finding bugs.

Edit: ok...   continuing.

belisoful:
@ctrlaltca  Now that the WSDL gen repo is fixed for namespaces providers. this is a go.  All new test classes need namespaces and so all WIP is affected.  I never like pushing changes like this in the short term, knowing its needed for the long term.

Issue #1133, opened 2026-05-24
🔴 Wrapping up v4.3.3 (Issues with SqlMap and MySql)


belisoful:
This needs to be fixed before we ship 4.3.3.  Given the importance of this class. This is one of the few exceptions to changing this late.  Other parts were fixed up in an earlier PR and this is finishing the fix on that bug.

https://github.com/pradosoft/prado/blob/469bf5c28df2c5e4d05de8255479188c5c64e8ca/framework/Prado.php#L432

```php
public static function using($namespace): ?string
	{
		$namespace = static::prado3NamespaceToPhpNamespace($namespace);

		if (isset(self::$_usings[$namespace])) {
			return $namespace . '\\';
		}
		if (class_exists($namespace, false) ||
			interface_exists($namespace, false) ||
			trait_exists($namespace, false)) {
			return $namespace;
		}
	...
```

needs to be this:
```php
	public static function using($namespace): ?string
	{
		$namespace = static::prado3NamespaceToPhpNamespace($namespace);

		if (isset(self::$_usings[$namespace])) {
			return $namespace . '\\';
		}
		if (class_exists($namespace, false) ||
			interface_exists($namespace, false) ||
			trait_exists($namespace, false)) {
			if (($shortPos = strrpos($namespace, '\\')) !== false) {
				$shortName = substr($namespace, $shortPos + 1);
				if (!class_exists($shortName, false) &&
					!interface_exists($shortName, false) &&
					!trait_exists($shortName, false)) {
					class_alias($namespace, $shortName);
				}
			}
			return $namespace;
		}
```

belisoful:
@ctrlaltca Given the direction of Prado v4.4 in isolating ActiveRecord and SqlMap in their own repos, The current state of v4.3.3 may be "reasonable."   Basic Data supports 3rd parties.  bringing 3rd party to AR and SqlMap is beyond 4.3.3 scope.

Before you start tagging the release, please let me know so I can scope things out in Data AS IS for a less aggressive 3rd party Data roll-out.  Let me know so I can audit the current state for the transition of AR and SqlMap to their own repo.  (meaning, 3rd party support for those subsystems will not be in 4.3.3).

Adding 3rd party db support in  AR and SqlMap added a lot of complexity, too much.

belisoful:
@ctrlaltca We should probably add `friendsofphp/php-cs-fixer` and `phpstan/phpstan` to the 'require-dev' of composer.json and remove it from the "devtools" script.

[...]

PR #1274, opened 2026-09-17
Two corrections to the PHPStan extensions

belisoful:
### `isa()` did not narrow intersection subjects

PHPStan hands a method call to a `MethodTypeSpecifyingExtension` only when the subject type has exactly one object class name (`MethodCallHandler::specifyTypes()` gates on `count($referencedClasses) === 1`). An intersection such as `IService&TComponent` has two, so `TComponentIsaTypeSpecifyingExtension` was never consulted — `isMethodSupported()` was never even called. Nothing inside the extension could fix this.

`TComponent::isa()` now carries the narrowing itself, which PHPStan evaluates for every subject type:

```php
/**
 * @template T of object
 * @param class-string<T>|T $class class or string
 * @phpstan-assert-if-true T $this
 */
```

The tag ships with the framework source and needs no `phpstan.neon` entry, so it also removes the need for an inline `@var` after an `isa()` guard.

Because the tag cannot be switched off by configuration, the two `testIsa*_FailsWithoutExtension` tests no longer had a valid premise and became `_NarrowsWithoutServicesConfig`. `IsaIntersectionNegativeFixture.php` replaces the lost signal: it calls a method that exists nowhere and asserts the error names the class the guard narrowed to, which fails if the narrowing regresses.

### Virtual properties narrowed under one spelling only

PHPStan asks for a property using the spelling written in the source, and PRADO writes `Title`. The guard extensions registered `HasPropertyType(lcfirst($name))` only, so:

| Guard | Access | Before |
|---|---|---|
| `canGetProperty('Title')` | `$c->Title` | undefined property |
| `canSetProperty('Title')` | `$c->Title = 'x'` | undefined property |
| `hasMethod('getTitle')` | `$c->Title` | undefined property |

All three extensions now register both spellings. The duplicated `getjs`/`setjs`/`get`/`set` suffix parsing is extracted as `TComponentHasMethodTypeSpecifyingExtension::propertyNameFromAccessor()`.

### Documented, not fixed

The single-object-class-name rule also blocks **union** subjects (`TControl|TStyle`), and it still applies to `hasMethod()`, `canGetProperty()` and `canSetProperty()`. Those narrow to `HasMethodType`/`HasPropertyType`, which no PHPDoc tag expresses, so the `isa()` approach is not available to them. `Prado::method_visible($obj, 'foo')` does work on those subjects — it is a *static* method extension, found through the `Prado` class and narrowing an argument, so the subject's class-name count never enters the lookup. This is now documented in both `INDEX.md` files and the four affected knowledge files, with a worked example.

### Tests

`phpunit --testsuite phpstan`: 19 → 21 tests, all passing. New `PropertyCaseFixture.php` (7 errors without the extensions, 0 with) and `IsaIntersectionFixture.php` / `IsaIntersectionNegativeFixture.php`; `MethodVisibleFixture.php` gains an intersection case.

`php -l`, `php-cs-fixer --dry-run framework/` (0 of 957) and `phpstan analyse framework/` (no errors) are clean. `phpunit --testsuite unit` has the usual 3 errors + 1 failure from the MySQL/PostgreSQL tests, unchanged by this branch — verified by re-running them with the changes reverted.


belisoful:
also, back port the PHPStan fixes to #1133.

belisoful:
This just fixed a regression in the composer.json.   The commit was added in the middle of creating the test PR, which it did not catch.

Issue #1006, opened 2025-11-04
LoginUser custom validation is not working on blog tutorial (v 4.3.1)


aurelio1811:
Hello, my name is Aurelio Martinez, from Venezuela (South America). first all all, sorry for my bad english because my native language is spanish.
 I am learning web development with php and prado framework, using php version 8.3.27 on linux mint 22.1 (xia), an prado 4.3.1.
 but I have problems with blog tutorial. I installed prado using: "composer create-project pradosoft/prado-app blog", and then: "composer update", adding  "pradosoft/prado-wsat": "^4" in the "require" section of composer.json. I am using mysql instead of sqlite as database.
     Problem arises when testing LoginUser page: custom validator not working. Any input in the Username and Password control always generate same exception: **"[Unknown Error] base64_decode(): Passing null to parameter #1 ($string) of type string is deprecated (@line 68 in file /var/www/html/blog/protected/vendor/pradosoft/prado/framework/Web/UI/TPageStateFormatter.php). (Error Code: 0)"**, as shown in attached file: "LoginUser_Exception.png".

 I attached some other files. Any help will be appreciated. Thanks.

<img width="1440" height="900" alt="Image" src="https://github.com/user-attachments/assets/7a888546-9440-44b8-96cb-4ed67ce4b6c7" />

[application.xml](https://github.com/user-attachments/files/23346293/application.xml)

[other_files.zip](https://github.com/user-attachments/files/23346501/other_files.zip)

ctrlaltca:
Yep, the underlying problem is probably a corrupted page state. 
This should never happen on a production website, but it may be caused by some operations during development
You can either try to:
 * delete the contents of protected/runtime/application.xml-4.3.1, so that the app starts on a fresh state
 * clean your browser cache or browse the page in Incognito mode, so it won't reuse the old "broken" session

aurelio1811:
I deleted the contents of protected/runtime/application.xml-4.3.1 and cleaned your browser cache, and problems persists (Bad Request - Page state is corrupted).

belisoful:
C4. TPageStateFormatter::unserialize() — null page state raises a base64_decode() deprecation instead of the "page state corrupted" error

Files: framework/Web/UI/TPageStateFormatter.php:68 (../prado.prado-4.3)
```php
public static function unserialize($page, $data)
{
    $str = base64_decode($data);   // line 68 — $data may be null
```
[...]

PR #1277, opened 2026-09-18
Fixed: TPageStateFormatter::unserialize() deprecation on a missing page state

belisoful:
A postback carrying `PRADO_POSTBACK_TARGET` without `PRADO_PAGESTATE` reaches
`TPageStateFormatter::unserialize()` with null, and `base64_decode()` was called
before the empty check. PHP 8.1 emits "Passing null to parameter #1 ($string) of
type string is deprecated", which Debug mode surfaces as an `[Unknown Error]`
naming neither the page state nor the problem. Reported in #1006.

- Guard a null or empty `$data` before `base64_decode()`, and check the decode
  result in the same guard. The persisters already turn the null return into the
  `THttpException` 400 "page state corrupted".
- Add `TPageStateFormatterTest` (the class had none): the missing-state cases
  assert no PHP diagnostic is raised, plus round trips and tamper rejection for
  validation, compression, encryption, IGBinary, and the default page settings.
- Record the missing-state contract in `agents/framework/Web/UI/TPageStateFormatter.md`.

Backward compatible; no signature or message-file changes.


belisoful:
Half the fix for #1006 is in prado-demos.  This PR fixes the second half (in Prado) of the issue.

